PR #118, opened 2026-01-16
Add media batch handling for Telegram messages

### LINKED ISSUE #106: Telegram: Support multi-image and file uploads
Before the agent starts its turn, it should have received all images and files that the user has provided. For example the user attaches 4 images, the agent should get all 4 images as part of the user prompt. Similarly when the user provides files, the files should exist first before we give the agent the injected context and user prompt so that it can access all files at once, instead of multiple files kicking off multiple turns

## Goal

When a Telegram user sends multiple images and/or files in one “send” action (e.g., an album), the bot should start **one** agent turn that includes **all** attachments. No intermediate agent turns should start until all attachments in that batch are downloaded/saved and ready for the agent to access.

## Telegram Bot API constraints to leverage

* Telegram includes `media_group_id` on each `Message` that belongs to a media group (album). This is the primary signal to group multi-image uploads. ([[Telegram](https://core.telegram.org/bots/api)][1])
* Telegram’s `sendMediaGroup` describes albums as containing 2–10 items and may include photos/videos/documents/audios; documents and audios are restricted to albums with the same type. This is useful for anticipating batch sizes and mixed-media behavior. ([[Telegram](https://core.telegram.org/bots/api)][1])
* File download flow is `getFile -> file_path -> download`. The `File` object documents that the download URL is derived from `file_path` and is time-limited, which matches the existing implementation pattern (`get_file` then `download_file`). ([[Telegram](https://core.telegram.org/bots/api)][1])

## Current behavior in this repo (why it multi-turns today)

In `src/codex_autorunner/integrations/telegram/handlers/messages.py`:

* `handle_message()` sets `bypass = has_media`, so **every** media message immediately flushes text coalescing and calls `handle_message_inner()`.
* `handle_message_inner()` routes `has_media` to `handle_media_message()`, which ends up calling `_handle_image_message()` / `_handle_file_message()` in `handlers/commands_runtime.py`.
* Those functions download/save the one attachment and then call `_handle_normal_message()` — which starts an agent turn.
* Therefore, “4 images” → 4 Telegram messages → 4 agent turns.

## Target behavior

For a burst of media messages from the same user in the same topic:

1. Buffer them into a “pending upload batch”.
2. Wait briefly (debounce) to collect the full set.
3. Download/save **all** attachments first.
4. Start **one** `_handle_normal_message()` call with:

   * `input_items`: one `"text"` item + **N** `"localImage"` items
   * prompt text that lists saved file paths for uploaded documents (and optionally describes images count)
5. Only after that turn completes (and outbox sending runs) should subsequent user inputs create additional turns.

### 1) Add `media_group_id` to the parsed message model

**Files**

* `src/codex_autorunner/integrations/telegram/adapter.py`
* `tests/test_telegram_adapter.py`

**Changes**

* Extend `TelegramMessage` dataclass with:

  * `media_group_id: Optional[str] = None`
* In `_parse_message()`, parse:

  * `media_group_id = payload.get("media_group_id")`
  * keep it only if it’s a non-empty `str`
* Add/extend a unit test in `test_parse_update_message()` to include `"media_group_id": "abc123"` and assert it round-trips.

Rationale: grouping albums requires this field. ([[Telegram](https://core.telegram.org/bots/api)][1])

---

### 2) Introduce a “media batch buffer” alongside the existing text coalescer

**Files**

* `src/codex_autorunner/integrations/telegram/service.py`
* `src/codex_autorunner/integrations/telegram/handlers/messages.py`

**Data structure**
Create a new buffer type in `handlers/messages.py`, similar to `_CoalescedBuffer`:

* `_MediaBatchBuffer`

  * `topic_key: str`
  * `messages: list[TelegramMessage]`  (photos/documents only)
  * `task: Optional[asyncio.Task[None]]`
  * optionally track:

    * `media_group_id: Optional[str]`
    * `created_at` / `last_updated_at` (for logging)

**Keys**
You need two levels of grouping:

1. Strong grouping when Telegram provides `media_group_id`:

* `batch_key = f"{topic_key}:user:{from_user_id}:mg:{media_group_id}"`

2. Fallback grouping when there is no `media_group_id` (for multi-file sends that come as separate messages):

* `batch_key = f"{topic_key}:user:{from_user_id}:burst"`

This fallback is what prevents “multi-file selection” from producing multiple turns if Telegram doesn’t set `media_group_id`.

**Debounce window**

* Add a new constant (or config) for media batching, e.g. `MEDIA_BATCH_WINDOW_SECONDS = 1.0` (or reuse 2.0).
* Each time a new eligible media message arrives, cancel and reschedule the flush task for that key.

**Service state**
In `TelegramBotService.__init__` (service.py), add:

* `self._media_batch_buffers: dict[str, _MediaBatchBuffer] = {}`
* `self._media_batch_locks: dict[str, asyncio.Lock] = {}`

Optionally add cache eviction support similar to `coalesced_buffers` (not strictly required because flush tasks are short-lived, but consistent with existing patterns).

---

### 3) Modify message routing so photos/documents are buffered instead of immediately starting turns

**File**

* `src/codex_autorunner/integrations/telegram/handlers/messages.py`

**Update `handle_message()` logic**
Currently:

* `bypass = has_media`

Change behavior to:

* Keep bypass for:

  * interrupts (`^C`, `/stop`, etc.)
  * shell bangs (`!cmd`)
  * recognized commands (including captions that are commands)
  * voice/audio (recommended non-goal for this feature unless explicitly required)

* For **batchable media** (photos and documents):

  * flush any pending text coalescing buffer for this user/topic (optional but preserves current ordering semantics)
  * add the message to the media batch buffer
  * return without starting a turn

This ensures multiple photo/document messages get collected.

---

### 4) Add a single “process media batch” entrypoint that starts exactly one agent turn

**Files**

* `src/codex_autorunner/integrations/telegram/handlers/commands_runtime.py`
* (possibly) `src/codex_autorunner/integrations/telegram/handlers/messages.py`

**Why here**
The best reuse is in `TelegramCommandHandlers` (commands_runtime.py) because it already contains:

* `_download_telegram_file()`
* `_save_image_file()`
* `_save_inbox_file()`
* `_format_file_prompt()` / file hint templates
* `_handle_normal_message()` (the turn starter)

**New method (recommended)**
Add something like:

* `async def _handle_media_batch(self, messages: Sequence[TelegramMessage], runtime: Any) -> None:`

Responsibilities:

1. Resolve topic + bound workspace once

* mimic checks from `handle_media_message()`:

  * record exists
  * record.workspace_path exists
  * media config enabled

2. Extract candidates in message order

* For each message:

  * `select_image_candidate(message)` → list of image candidates
  * `select_file_candidate(message)` → list of file candidates
* Keep stable ordering (sort by `message_id`)

3. Download/save everything before starting the turn

* Images:

  * for each candidate:

    * `data, file_path, file_size = await self._download_telegram_file(file_id)`
    * enforce `max_image_bytes`
    * `saved = self._save_image_file(workspace_path, data, file_path, candidate)`
* Files:

  * similarly enforce `max_file_bytes`
  * `saved = self._save_inbox_file(workspace_path, topic_key, data, candidate=..., file_path=...)`

4. Build ONE combined prompt text
   Include:

* any captions present in the batch:

  * choose the first non-empty caption as the main instruction, or concatenate captions with separators
* a file summary section listing all saved paths (and optionally sizes/types)
* include the existing inbox/outbox hint **once** (reuse `FILES_HINT_TEMPLATE` with the topic dirs)

5. Build ONE `input_items`

* start with `{"type": "text", "text": combined_prompt}`
* then one `{"type": "localImage", "path": str(saved_image_path)}` per image
* (files remain referenced by path in the text; no extra input type needed given current architecture)

6. Start exactly one turn

* choose a reply target message id:

  * use the last message in the batch (`max(message_id)`) so the bot reply visually follows the batch
* call:

  * `await self._handle_normal_message(reply_message, runtime, text_override=combined_prompt, input_items=input_items, record=record)`

This meets the requirement that files “exist first” and all images are provided to the agent in the same request.

---

### 5) Wire the buffer flush to enqueue one batch-processing job

**File**

* `src/codex_autorunner/integrations/telegram/handlers/messages.py`

When the debounce timer fires:

* Atomically pop the buffer (under lock)
* Build `topic_key` and `runtime`
* Enqueue one job on the topic queue:

  * `handlers._enqueue_topic_work(topic_key, lambda: handlers._handle_media_batch(buffer.messages, runtime))`

This keeps ordering consistent with per-topic queue behavior and avoids starting a batch turn concurrently with another turn.

---

### 6) Size limits, caps, and failure handling

Add explicit policies to avoid pathological batches:

* **Cap attachments per batch**:

  * default cap of 10 aligns with album semantics described by [Telegram](https://core.telegram.org/bots/api). ([Telegram][1])
  * If more arrive within the debounce window, either:

    * split into multiple turns (e.g., first 10 then next 10), or
    * reject extras with a clear message
* **Partial download failures**:

  * If some downloads fail:

    * proceed with the successfully saved attachments
    * include a “failed to download X items” line in the prompt
  * If all downloads fail:

    * send a Telegram error message and do not start a turn
* **Oversize handling**:

  * preserve existing behavior (send an error instead of downloading/processing)
  * enforce per-file max bytes using existing config fields

---

### 7) Tests to add

You’ll want tests at two levels: parsing + batching behavior.

1. **Adapter parse test**

* Extend `tests/test_telegram_adapter.py` to validate `media_group_id`.

2. **Batching behavior tests**
   Add a new async test module, e.g. `tests/test_telegram_media_batching.py`, that:

* builds a `TelegramBotService` with a fake bot + bound workspace
* stubs:

  * `_download_telegram_file()` to return deterministic bytes + paths
  * `_handle_normal_message()` to record invocations instead of running the app server
* sends 4 “photo messages” with the same `media_group_id` through `message_handlers.handle_message(service, msg)`
* advances the event loop past the debounce window (`await asyncio.sleep(window + epsilon)`)
* asserts:

  * `_handle_normal_message()` was called **once**
  * `input_items` contains 1 text + 4 localImage items

Repeat for 3 document messages (files) and assert:

* files exist on disk before `_handle_normal_message()` is invoked
* prompt includes all saved file paths

This directly validates the feature requirement.

---

### 8) Configuration and rollout

To minimize risk, add a config flag under `TelegramBotMediaConfig`, e.g.:

* `batch_uploads: bool` (default `True` or `False` depending on desired backward-compat)
* `batch_window_seconds: float` (default `1.0`–`2.0`)

Wire it so:

* if disabled → current behavior (each media message starts its own turn)
* if enabled → buffering behavior

---

## Non-goals (keep scope tight)

* Voice/audio batching (voice notes already have their own workflow and progress messages).
* Video batching (this integration doesn’t currently parse `video` fields).
* Merging text-only coalescing with media batching (optional enhancement later).

This plan delivers multi-image and multi-file uploads in [Telegram](https://core.telegram.org/bots/api) as a single agent turn, using `media_group_id` where available and a debounce fallback to cover multi-file sends that arrive as separate messages. ([Telegram][1])

[1]: https://core.telegram.org/bots/api "https://core.telegram.org/bots/api"

Git-on-my-level:
## Summary

This PR implements media batch handling for Telegram messages, allowing multiple media items to be processed together.

## Changes

- **adapter.py**: Added  field to  for grouping related media from Telegram
- **config.py**: Added  and  configuration options for media handling
- **constants.py**: Added  constant
- **handlers/messages.py**: Implemented buffering and flushing logic for media batches with proper async locking
- **handlers/commands_runtime.py**: Added  method to process batched media items
- **service.py**: Added  and  dictionaries with cache eviction support
- **tests**: Updated and added tests for new media batch functionality

## Key Features

1. **Media Grouping**: Messages with the same  from Telegram are batched together
2. **Burst Mode**: Without , messages from the same user/topic are batched within a time window
3. **Configurable**: Batch behavior can be disabled or tuned via configuration
4. **Proper Locking**: Async locks prevent race conditions during buffer updates
5. **Cache Eviction**: Old batches are cleaned up after TTL expires

## Fixes

Fixes #106


Git-on-my-level:
@codex review

Git-on-my-level:
@codex review

Git-on-my-level:
@codex review

PR #144, opened 2026-01-17
Make agent binaries config-driven

Git-on-my-level:
## Summary
- add agents config map with binary/serve_command defaults
- plumb opencode binary/serve command through config in terminal, Telegram, and app server
- document agent settings in codex-autorunner.yml and annotate agent types

## Testing
- black
- ruff
- mypy
- eslint
- tsc
- pytest

### DIFF
--- a/codex-autorunner.yml
+++ b/codex-autorunner.yml
@@ -45,13 +45,14 @@ repo:
       binary: codex
     opencode:
       binary: opencode
-      serve_command:
-        - opencode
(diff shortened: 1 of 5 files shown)


Git-on-my-level:
Resolved review: opencode supervisor now derives default serve command from configured binary, falling back only when explicit serve_command provided. Also removed hardcoded serve_command defaults from config + docs.

Issue #556, opened 2026-02-06
PMA ergonomics: dispatch reply, archive candidate tooling, manifest repair, and hub capability checks


Git-on-my-level:
## Summary
PMA coordination surfaced several recurring friction points that slow down routine operations and increase operator error risk. This issue proposes targeted CAR improvements so common PMA workflows are one-command, self-validating, and easier to reason about.

Date observed: 2026-02-06
Context: Hub-level PMA operation across many repos/worktrees, including dispatch handling and bulk worktree archival.

### 1) No PMA-native dispatch reply command
**Current behavior**
- Paused runs appear in hub snapshot / Inbox.
- PMA can read the dispatch prompt and propose an answer, but cannot post the answer + resume in one CLI-native step.
- This forces manual Inbox UI interaction, which breaks automation and slows triage.

**Proposed implementation**
- Add command: `car hub dispatch reply` (or `car pma dispatch reply`) with:
  - `--repo-id`
  - `--run-id`
  - `--message` or `--message-file`
  - optional `--resume` (default true)
- Behavior:
  - Persist reply into dispatch history in the same canonical location used by UI flow.
  - Resume paused run if `--resume` is true.
  - Return machine-readable JSON output with posted `dispatch_seq`, status, and resumed run id.
- Add guardrails:
  - Reject if run is not paused-awaiting-input.
  - Idempotency key option to prevent duplicate replies on retries.

**Acceptance criteria**
- A paused run can be answered and resumed without visiting Inbox tab.
- Command exits non-zero on invalid run state with actionable error.
- JSON output supports scripting in PMA workflows.

---

### 2) No unified archive-candidate view
**Current behavior**
- PMA must infer archival eligibility from mixed signals (`ticket_flow`, repo status, relevance context).
- There is no first-class "safe to archive" report in one place.

**Proposed implementation**
- Add command: `car hub worktree candidates --archive --json`
- Output fields per worktree:
  - `repo_id`, `worktree_of`, `branch`, `status`
  - `ticket_flow_status`, `done_count`, `total_count`
  - `has_paused_dispatch`, `has_active_run`
  - `last_activity_at`, `recommended_action`, `reasons[]`
- Include policy modes:
  - `strict`: only fully completed/no active blockers
  - `lenient`: includes stale/idle with warning labels

**Acceptance criteria**
- PMA can list and explain archive candidates without manual cross-referencing tabs/files.
- Output includes explicit reasons so operators can trust automation.

---

[...]

PR #557, opened 2026-02-06
Add hub dispatch reply CLI command for PMA workflows

Git-on-my-level:
## Summary
This PR implements a first PMA ergonomics slice for #556: a hub-level dispatch reply command that can post a reply and resume in one step.

### Added
- `car hub dispatch reply`
- Options:
  - `--repo-id`
  - `--run-id`
  - `--message` or `--message-file` (exactly one required)
  - `--resume/--no-resume` (default: resume)
  - `--idempotency-key` (optional duplicate guard)
  - `--json/--no-json` and `--pretty`

### Behavior
- Validates run is paused before posting.
- Posts reply via hub-mounted repo message route.
- Resumes run in same command by default.
- Emits machine-readable JSON with `reply_seq`, `duplicate`, `resumed`, `resume_status`.
- If `--idempotency-key` is provided, it detects previously-posted replies for that key and avoids duplicate posting.

## Tests
- Added `tests/test_cli_hub_dispatch_reply.py`:
  - posts + resumes happy path
  - idempotency duplicate detection path
  - input validation for message source
- Existing `tests/test_cli_hub_snapshot.py` still passes.

## Validation
- Pre-commit hook suite executed on commit (black, ruff, mypy, eslint, static build, full pytest)
- Focused tests also run manually:
  - `.venv/bin/python -m pytest tests/test_cli_hub_dispatch_reply.py tests/test_cli_hub_snapshot.py`

Refs #556

### DIFF
--- a/src/codex_autorunner/surfaces/cli/cli.py
+++ b/src/codex_autorunner/surfaces/cli/cli.py
@@ -438,12 +438,21 @@ def _request_form_json(
     url: str,
     form: Optional[dict] = None,
     token_env: Optional[str] = None,
+    *,
+    force_multipart: bool = False,
 ) -> dict:
     headers = None
     if token_env:
         token = _require_auth_token(token_env)
         headers = {"Authorization": f"Bearer {token}"}
-    response = httpx.request(method, url, data=form, timeout=5.0, headers=headers)
+    data = form
+    files = None
+    if force_multipart:
+        data = form or {}
+        files = []
+    response = httpx.request(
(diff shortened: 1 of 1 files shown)


Git-on-my-level:
Addressed feedback: reply posting now forces multipart form data when calling /api/messages/{run_id}/reply (FastAPI expects File(...) so non-multipart could 422). Implemented via _request_form_json(force_multipart=True), and the hub dispatch reply now uses it.

PR #562, opened 2026-02-06
Fix path traversal vulnerabilities in Telegram uploads, file-write approval, and workspace resolution

mgpai22:
## Summary

Fixes three path traversal vulnerabilities:

- **Telegram attachment filename injection** (`service.py`): User-supplied filenames from Telegram attachments were joined directly with the reply directory. A crafted filename like `../../etc/cron.d/backdoor` could write to arbitrary paths. Fixed by stripping directory components via `Path(filename).name`.

- **File-write allowlist bypass** (`app.py`): The `_path_is_allowed_for_file_write` check used raw string prefix matching without collapsing `..` segments. A path like `.codex-autorunner/workspace/../../etc/passwd` would pass the prefix check but resolve outside the allowed directory. Fixed by normalizing with `posixpath.normpath` and rejecting paths that escape the prefix.

- **Workspace root escape** (`app.py`): `_resolve_workspace_and_runs` resolved a user-supplied workspace root relative to repo root but never verified the result stayed within the repo boundary. Fixed by adding a containment check after resolution.

## Test plan

- [ ] Verify Telegram file replies still write correctly with normal filenames
- [ ] Verify filenames containing `/` or `..` are stripped to their base name
- [ ] Verify file-write approval rejects paths with `..` segments that escape the allowlist
- [ ] Verify workspace root values that traverse above repo root raise `ValueError`
- [ ] Run existing test suite (`pytest`)

### DIFF
--- a/src/codex_autorunner/surfaces/web/app.py
+++ b/src/codex_autorunner/surfaces/web/app.py
@@ -217,9 +217,7 @@ def _resolve_workspace_and_runs(
         workspace_root == resolved_repo
         or str(workspace_root).startswith(str(resolved_repo) + os.sep)
     ):
-        raise ValueError(
-            f"workspace_root escapes repo boundary: {workspace_root}"
-        )
+        raise ValueError(f"workspace_root escapes repo boundary: {workspace_root}")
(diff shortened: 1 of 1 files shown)


mgpai22:
@codex review

Issue #618, opened 2026-02-12
Bug: Agent renaming current ticket file causes CAR workflow to pause


snavarro89:
## Bug Report

  When an AI agent with file system access renames the current ticket file (e.g., `TICKET-009.md` →
  `TICKET-009-ALREADY-DONE.md`), the CAR workflow pauses with error.

  The user must manually rename the file back (e.g., remove `-ALREADY-DONE`) to resume processing.

  ## Steps to Reproduce

  1. Create a ticket `TICKET-009.md` with `done: false`
  2. Start CAR flow - it processes TICKET-009
  3. Agent renames the file to `TICKET-009-ALREADY-DONE.md` (e.g., as a backup/marker) -> (I had no instruction on my agent to rename the file, not sure how or why it decided to rename it, other tickets I have didn't rename after they were done.)
  4. On next step, CAR pauses with error: "TICKET-009.md could not be found"

  ## Expected Behavior

  CAR should detect that the original file no longer exists and clear the stale path, then find the renamed file
  (`TICKET-009-ALREADY-DONE.md`) via `list_ticket_paths()` and continue processing.

  ## Actual Behavior

  CAR stores the original path in `state["current_ticket"]` and reconstructs `current_path` from this stale state on every
  step. The code doesn't check if the file still exists, so it keeps trying to read the non-existent file. After
  `max_lint_retries` is exceeded, the run pauses.

  ## Root Cause

  In `src/codex_autorunner/tickets/runner.py:292-297`:

  ```python
  current_ticket = state.get("current_ticket")
  current_path: Optional[Path] = (
      (self._workspace_root / current_ticket)
      if isinstance(current_ticket, str) and current_ticket
      else None
  )

  There's no check for whether current_path still exists on disk. When the agent renames the file, the stored path becomes
  stale but is never cleared.

  Later, ticket_is_done(current_path) returns False for the non-existent file (OSError is caught), so the stale path
  persists through max_lint_retries until the run pauses.

  Proposed Solution

  Add a check after line 297 in src/codex_autorunner/tickets/runner.py to detect when the current ticket file no longer
  exists and clear the stale path:

  # Detect if current ticket file was renamed/deleted by agent
  if current_path and not current_path.exists():
      _logger.warning(
          "Current ticket file no longer exists at %s (may have been renamed by agent). "
          "Clearing stale path and finding next ticket.",
          safe_relpath(current_path, self._workspace_root)
      )
      current_path = None
[...]

PR #633, opened 2026-02-13
Handle stale current_ticket when ticket file is renamed

Git-on-my-level:
## Summary\n- detect when persisted `state["current_ticket"]` points to a ticket path that no longer exists\n- clear stale per-ticket state and reselect the next open ticket from current on-disk ticket paths\n- add a regression test that reproduces the rename scenario and verifies runner recovery\n\n## Validation\n- .venv/bin/python -m pytest tests/tickets/test_ticket_runner.py -k stale -q\n- .venv/bin/python -m pytest tests/tickets/test_ticket_runner.py -q\n- .venv/bin/python -m pytest tests/tickets/test_runner_stale_pause.py -q\n- pre-commit/commit hooks ran full project checks and tests\n\nCloses #618

### DIFF
--- a/src/codex_autorunner/tickets/runner.py
+++ b/src/codex_autorunner/tickets/runner.py
@@ -310,6 +310,8 @@ class TicketRunner:
             state.pop("last_agent_output", None)
             state.pop("lint", None)
             state.pop("commit", None)
+            commit_pending = False
+            commit_retries = 0
 
         # If current ticket is done, clear it unless we're in the middle of a
         # bounded "commit required" follow-up loop.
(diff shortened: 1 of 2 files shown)


Git-on-my-level:
Follow-up (formatted correctly):

- reset in-memory `commit_pending` / `commit_retries` when clearing a missing stale `current_ticket` path (`src/codex_autorunner/tickets/runner.py`)
- strengthened rename-recovery regression test to assert commit-retry prompt does **not** leak to the newly selected ticket (`tests/tickets/test_ticket_runner.py`)

This prevents a new ticket from being treated as commit-retry mode after stale-path recovery.

PR #672, opened 2026-02-15
Fix: Classify SIGTERM worker exits as STOPPED instead of FAILED

### LINKED ISSUE #669: Flow worker SIGTERM exits still classify as failed in ticket_flow resume
## Summary
Flow worker `d44e3100-d421-47c0-b673-33516a9b54e8` repeatedly fails with a SIGTERM-based worker death while parked in `ticket_turn`, even after adding graceful shutdown changes to `flow_worker`.

## Environment
- Repo: `codex-autorunner--process-opencode-leak-remediation`
- Run id: `d44e3100-d421-47c0-b673-33516a9b54e8`
- Command retried: `/Users/dazheng/car-workspace/car flow ticket_flow resume --repo /Users/dazheng/car-workspace/worktrees/codex-autorunner--process-opencode-leak-remediation --run-id d44e3100-d421-47c0-b673-33516a9b54e8 --force`

## Reproduction Observed
1. Resume blocked run.
2. Flow enters `ticket_turn` on `TICKET-005-flow-worker-graceful-sigterm.md`.
3. `flow status` reports running during execution.
4. Worker exits as dead with `SIG15` and run transitions to `failed`.

### Run status
`car flow ticket_flow status --repo ... --run-id d44...` (immediately after resume):
- `Status: running`
- `Tickets: 1/11`
- `Current step: ticket_turn`
- `Current ticket: .codex-autorunner/tickets/TICKET-005-flow-worker-graceful-sigterm.md`
- `Worker: dead pid=19872 worker PID not running` (per status output and logs)

### flow event tail (from `flows.db`)
- `seq 4316` `flow_failed`: `{"error": "Worker died (status=dead, pid=19872, reason: worker PID not running, exit_code=-15)", "reason": "worker-dead", ...}`
- `seq 4315` `flow_failed`: same error with worker pid 95389
- `seq 4314` last in-process event: `app_server_event` (`account/rateLimits/updated`)

### Worker artifacts
- `.codex-autorunner/flows/d44e3100-d421-47c0-b673-33516a9b54e8/worker.exit.json`
  - `returncode: -15`
  - `pid: 95389`
  - no stdout/stderr tails (empty)
- `.codex-autorunner/flows/d44e.../crash.json`
  - `exit_code: -15`
  - `signal: SIG15`
  - `last_event: account/rateLimits/updated`
  - `stack_trace: null`, `exception: null`

### Additional runtime evidence
- `codex-server.log` includes reconciliation lines: `Reconciling flow ...: running -> failed (worker-dead)`.
- No `worker.err.log` or `worker.out.log` output.

### Source under test
- `src/codex_autorunner/surfaces/cli/commands/flow.py` has updated `flow_worker` handler:
  - SIGTERM/SIGINT handler currently sets `shutdown_event` only.
  - waits on `{run_task, shutdown_wait_task}` with `FIRST_COMPLETED`.
  - on shutdown requests it calls `await controller.stop_flow(run_id)` and `run_task.cancel()`.

## Hypothesis
The worker appears to be receiving SIGTERM from outside the flow runtime while the ticket agent may be mid-turn, but reconciliation still resolves the run as `RUNNING -> FAILED` (worker-dead) because transition logic treats any dead worker in `RUNNING` as failure; there is no visible stop intent to classify this as expected stop.

Given the new handler, we may still be missing an explicit persistence flag that a shutdown was requested. Without that, a legitimate SIGTERM path can race with transition evaluation and be interpreted as worker crash.

## Suggested fix
1. On signal, persist a shutdown intent before waiting (e.g., in run status/state or explicit metadata).
2. Update transition/reconcile path so a known “shutdown requested” shutdown from flow worker is mapped to `STOPPED`/`STOPPING` rather than `FAILED`.
3. Add regression coverage for SIGTERM arriving during active worker execution where the worker exits without explicit exception and is classified as stopped.
4. Capture the shutdown intent in failure artifacts/`flow_runs.state` for diagnostics parity with crash cases.

Git-on-my-level:
## Summary

Fixes #669

When a flow worker receives SIGTERM/SIGINT, it now persists a `shutdown_intent` flag in `worker.exit.json`. The transition logic checks this flag and transitions the run to STOPPED instead of FAILED, distinguishing graceful shutdown from unexpected crashes.

This fixes `ticket_flow resume` incorrectly classifying SIGTERM-based worker deaths as failures when the worker was parked in `ticket_turn`.

## Changes

1. **Added `shutdown_intent` parameter to `write_worker_exit_info`** (`worker_process.py:110-127`)
   - New boolean parameter defaults to `False`
   - Persisted in `worker.exit.json` for postmortem analysis

2. **Added `shutdown_intent` field to `FlowWorkerHealth`** (`worker_process.py:22-37`)
   - Exposed to transition logic via health check

3. **Updated `check_worker_health` to read `shutdown_intent`** (`worker_process.py:389-438`)
   - Reads from exit info when worker is dead
   - Only trusts exit info that matches current worker lifecycle

4. **Updated `resolve_flow_transition` for shutdown intent** (`transition.py:61-83`)
   - RUNNING + dead worker + `shutdown_intent=True` → STOPPED
   - RUNNING + dead worker + `shutdown_intent=False` → FAILED (unchanged)

5. **Updated signal handler in `flow.py`** (`flow.py:513-530`)
   - Sets `shutdown_intent=True` when writing exit info on SIGTERM/SIGINT

## Tests

- `test_shutdown_intent_transitions_to_stopped_not_failed`: Verifies transition to STOPPED
- `test_no_shutdown_intent_still_fails`: Verifies FAILED still applies without intent
- `test_write_worker_exit_info_with_shutdown_intent`: Verifies persistence
- `test_check_worker_health_reads_shutdown_intent`: Verifies health check reads intent

Closes #669

### DIFF
--- a/src/codex_autorunner/core/flows/worker_process.py
+++ b/src/codex_autorunner/core/flows/worker_process.py
@@ -137,10 +137,21 @@ def write_worker_exit_info(
         )
     except Exception:
         metadata = {}
+
+    existing_shutdown_intent = False
+    exit_path = _worker_exit_path(artifacts_dir)
+    if not shutdown_intent and exit_path.exists():
+        try:
+            existing = json.loads(exit_path.read_text(encoding="utf-8"))
+            if isinstance(existing, dict) and existing.get("shutdown_intent") is True:
+                existing_shutdown_intent = True
+        except Exception:
+            pass
+
     data = {
         "run_id": normalized_run_id,
         "returncode": returncode,
-        "shutdown_intent": shutdown_intent,
+        "shutdown_intent": shutdown_intent or existing_shutdown_intent,
         "captured_at": time.time(),
         "pid": metadata.get("pid"),
         "spawned_at": metadata.get("spawned_at"),
@@ -148,9 +159,7 @@ def write_worker_exit_info(
         "stdout_tail": _tail_file(artifacts_dir / "worker.out.log"),
     }
     try:
-        _worker_exit_path(artifacts_dir).write_text(
-            json.dumps(data, indent=2), encoding="utf-8"
-        )
+        exit_path.write_text(json.dumps(data, indent=2), encoding="utf-8")
     except Exception:
         pass
 
--- a/tests/flows/test_worker_crash.py
+++ b/tests/flows/test_worker_crash.py
@@ -234,3 +234,43 @@ def test_check_worker_health_reads_shutdown_intent(monkeypatch, tmp_path: Path) 
     assert health.status == "dead"
     assert health.shutdown_intent is True
     assert health.exit_code == -15
+
+
+def test_write_worker_exit_info_preserves_existing_shutdown_intent(
+    tmp_path: Path,
+) -> None:
+    repo_root = tmp_path
+    run_id = "723e4567-e89b-12d3-a456-426614174000"
+
+    artifacts_dir = repo_root / ".codex-autorunner" / "flows" / run_id
+    artifacts_dir.mkdir(parents=True, exist_ok=True)
+
+    from codex_autorunner.core.flows.worker_process import (
+        _worker_metadata_path,
+        _write_worker_metadata,
+    )
+
+    metadata_path = _worker_metadata_path(artifacts_dir)
+    _write_worker_metadata(metadata_path, 12347, ["python", "-m", "test"], repo_root)
+
+    write_worker_exit_info(
+        repo_root,
+        run_id,
+        returncode=-15,
+        shutdown_intent=True,
+    )
+
+    exit_path = artifacts_dir / "worker.exit.json"
+    assert exit_path.exists()
+    payload = json.loads(exit_path.read_text(encoding="utf-8"))
+    assert payload["shutdown_intent"] is True
+
+    write_worker_exit_info(
+        repo_root,
+        run_id,
+        returncode=-15,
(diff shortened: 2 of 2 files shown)


Git-on-my-level:
## Implementation Notes

This fix addresses the root cause identified in #669: when a flow worker receives SIGTERM while parked in `ticket_turn`, the transition logic had no way to distinguish an intentional shutdown from an unexpected crash.

### Key Design Decision

Rather than modifying the existing `stop_requested` flag (which requires the controller to set it), I added a new `shutdown_intent` flag that the **worker itself** persists before dying. This works because:

1. The signal handler runs before the process terminates
2. It writes `shutdown_intent=True` to `worker.exit.json`
3. The reconciler reads this flag during health checks
4. Transition logic uses it to map `RUNNING → STOPPED` instead of `RUNNING → FAILED`

This approach is resilient to external SIGTERM sources (e.g., system shutdown, process managers) that don't go through the CAR controller's `stop_flow()` method.

### Why not just check for exit_code=-15?

Exit codes alone are ambiguous - a SIGTERM could indicate:
- Graceful shutdown (intended)
- OOM killer (not intended)
- Resource limits (not intended)

The `shutdown_intent` flag explicitly captures the worker's **intent** at signal-receipt time, making the semantics clear.

PR #799, opened 2026-03-01
Unify Discord/Telegram audio inference and fix remaining Discord Whisper gap

Git-on-my-level:
## Summary
This PR addresses the remaining Discord Whisper hookup gaps and reduces divergence with Telegram by centralizing media/audio inference logic.

### What changed
- Added shared media helpers in `integrations/chat/media.py`:
  - `normalize_mime_type`
  - `is_audio_mime_or_path`
  - `audio_content_type_for_input`
  - `audio_extension_for_input`
- Rewired **Discord adapter/service** to use shared helpers:
  - Detect audio even when `content_type` is generic (e.g. `application/octet-stream`) by using fallback signals like duration/suffix.
  - Normalize transcription `content_type` before Whisper calls.
  - Infer audio filename suffixes centrally and apply a safe `.ogg` fallback for audio-classified attachments without suffix.
- Rewired **Telegram message/voice path** to use the same shared helpers:
  - Normalize/infer audio MIME in `select_voice_candidate`.
  - Use shared extension inference for persisted voice files.
  - Normalize transcription MIME before Whisper calls.

## Root cause addressed
A remaining Discord case was still unhandled: when `content_type` existed but was generic/non-audio (instead of missing), adapter inference did not classify the attachment as audio, so transcription was skipped entirely. This also kept MIME handling inconsistent with Telegram.

## Tests
Added and updated regression coverage:
- `tests/test_chat_media.py` (new shared-helper tests)
- `tests/integrations/discord/test_adapter.py`
  - generic MIME + duration now infers `audio`
- `tests/integrations/discord/test_message_turns.py`
  - missing MIME now normalizes to audio content type for transcription
  - generic MIME + duration transcription path
- `tests/test_telegram_handlers_messages.py`
  - generic MIME normalized by suffix in voice candidate selection

Local validation:
- full pre-commit suite passed (including mypy, lint, static build, full pytest)

### DIFF
--- a/src/codex_autorunner/integrations/discord/service.py
+++ b/src/codex_autorunner/integrations/discord/service.py
@@ -1202,12 +1202,13 @@ class DiscordBotService:
                 source_url=source_url if isinstance(source_url, str) else None,
                 kind=getattr(attachment, "kind", None),
             )
-            suffix = audio_extension_for_input(
-                mime_type=getattr(attachment, "mime_type", None),
-                file_name=getattr(attachment, "file_name", None),
-                source_url=source_url if isinstance(source_url, str) else None,
-                default=".ogg" if is_audio else "",
-            )
+            if is_audio:
+                suffix = audio_extension_for_input(
+                    mime_type=getattr(attachment, "mime_type", None),
+                    file_name=getattr(attachment, "file_name", None),
+                    source_url=source_url if isinstance(source_url, str) else None,
+                    default=".ogg",
+                )
         return f"{stem[:64]}-{uuid.uuid4().hex[:8]}{suffix}"
 
     async def _find_paused_flow_run(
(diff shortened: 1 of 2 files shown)


Git-on-my-level:
Addressed the review comment in commit `84e99957`.

What changed:
- In `_build_attachment_filename`, audio extension inference now runs only when the attachment is already classified as audio.
- Non-audio items (including `video/mp4`) no longer flow through `audio_extension_for_input(...)`, so they cannot be mislabeled as `.m4a`.
- Added regression coverage: `test_build_attachment_filename_does_not_infer_audio_suffix_for_video` in `tests/integrations/discord/test_message_turns.py`.

Validation:
- Targeted tests for the filename/video path pass.
- Full hook suite passed (lint, typecheck, full pytest).

PR #972, opened 2026-03-15
Propagate ZeroClaw config dir for managed workspaces

Git-on-my-level:
## Summary
- propagate `ZEROCLAW_CONFIG_DIR` alongside `ZEROCLAW_WORKSPACE` for managed ZeroClaw launches
- reuse a shared managed-workspace env helper for both direct and Docker-wrapped launches
- document the canonical config-dir behavior and stop implying any install path is valid unless the CLI advertises `--session-state-file`

Follow-up to #969.

## Testing
- `PYTHONPATH=src .venv/bin/pytest -q tests/agents/zeroclaw/test_zeroclaw_supervisor.py`
- `ZEROCLAW_DOGFOOD=1 ZEROCLAW_TEST_MODEL=zai/glm-5 PYTHONPATH=src .venv/bin/pytest -q tests/agents/zeroclaw/test_zeroclaw_host_integration.py -m integration`
- commit hooks: `2837 passed, 1 skipped`

### DIFF
--- a/src/codex_autorunner/agents/zeroclaw/supervisor.py
+++ b/src/codex_autorunner/agents/zeroclaw/supervisor.py
@@ -1,8 +1,10 @@
 from __future__ import annotations
 
 import asyncio
+import fnmatch
 import json
 import logging
+import os
 import re
 import time
 import uuid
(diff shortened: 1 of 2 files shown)


Git-on-my-level:
Addressed the review in 734e9976. Docker-wrapped ZeroClaw launches now only inject the default ZEROCLAW_CONFIG_DIR when the destination is not already supplying that key explicitly or via env_passthrough, and I added regression coverage for both cases.

PR #980, opened 2026-03-16
Fix PMA stale state, reusable-thread status, and prompt delta injection

### LINKED ISSUE #975: PMA dogfooding feedback: setup ergonomics, stale state, and context/token efficiency
## Summary

Dogfooding PMA surfaced a few UX and state-model gaps that made the experience less ergonomic and less efficient than it should be.

## What worked

- The core path is conceptually sound: create agent workspace, verify runtime readiness, spawn PMA managed thread, send a task.
- ZeroClaw runtime readiness is explicit and testable.
- Managed threads are the right abstraction for exploratory PM work.

### 1. Stale inbox cards / dispatch text still undermine operator trust

I still had to mentally reconcile stale inbox cards against fresher run state and active-context notes. The docs already say to trust current run state over stale dispatch text, but needing to do that routinely is still a UX problem.

Implication:
- operators spend tokens and attention deciding which state is canonical

Suggested direction:
- stronger stale-state suppression in inbox
- clearer "superseded by newer run state" UI treatment
- one canonical status summary per run/thread

### 2. Managed thread statuses are technically correct but not intuitive

`completed lifecycle=active` is internally defensible, but it is not immediately obvious to an operator scanning thread state. I had to translate that into "the last turn completed; the thread is still reusable".

Suggested direction:
- present reusable thread state more directly, for example `idle/reusable` after a completed turn
- reserve `completed` for terminal thread lifecycle when that is truly the end of the thread

### 3. PMA entry paths are powerful but cognitively expensive

There are multiple valid control surfaces:
- paused run dispatches
- managed threads
- PMA file inbox
- ticket flows
- automation

That flexibility is useful, but it increases first-action ambiguity. The system expects the operator to know which lane should take precedence.

Suggested direction:
- stronger default routing in UI and CLI
- clearer "why this branch" explanation when PMA chooses dispatches vs files vs managed thread work

### 1. Repeated PMA doc injection is expensive

The PMA prompt, AGENTS content, active context, and hub snapshot are useful, but they are long and often repeated verbatim across turns. That increases cost and makes it harder to spot the genuinely new facts.

Suggested direction:
- stabilize the durable PMA prompt once, then send only deltas
- summarize active context aggressively and keep the full snapshot available by reference
- include "what changed since last turn" as a compact header

### 2. Large stale context competes with live state

A lot of the active context is historical reconciliation. It is useful for auditability, but it competes with current actionable state in the token budget.

Suggested direction:
- automatically compact superseded flow notes
- keep only currently actionable items in `active_context.md`
- move older reconciled items to `context_log.md` earlier

### 3. The agent had to infer precedence across sources

I had to combine:
- hub inbox state
- active_context notes
- run freshness timestamps
- managed thread state

This is workable, but not token efficient.

Suggested direction:
- emit a single normalized "action queue" for PMA with freshness, precedence, and supersession already resolved

## Concrete product opportunities

1. Add a canonical "freshest truth" layer so stale dispatch cards are visibly downgraded or hidden.
2. Rename or normalize reusable managed-thread statuses to reduce operator translation overhead.
3. Send PMA context as deltas/summaries instead of re-injecting the same long docs every turn.
4. Introduce a single PMA action queue ordered by precedence and freshness.

Git-on-my-level:
## Summary
- add a canonical PMA action queue and use it to downgrade superseded stale-state notifications
- introduce operator-facing reusable managed-thread status and roll it through CLI, PMA snapshot text, and web surfaces
- make PMA prompt assembly delta-aware so repeated turns stop re-injecting unchanged long context blocks by default
- update PMA docs and regression coverage for the new action model, status contract, and prompt behavior

## Testing
- Not run in this PR creation step

Closes #975

### DIFF
--- a/src/codex_autorunner/core/pma_context.py
+++ b/src/codex_autorunner/core/pma_context.py
@@ -309,27 +309,21 @@ def _build_prompt_bundle_digest(sections: Mapping[str, Mapping[str, Any]]) -> st
     return _digest_text(raw)
 
 
-def _load_pma_prompt_state(hub_root: Path) -> dict[str, Any]:
-    path = default_pma_prompt_state_path(hub_root)
-    lock_path = _prompt_state_lock_path(path)
-    with file_lock(lock_path):
-        if not path.exists():
-            return _default_pma_prompt_state()
-        try:
-            raw = path.read_text(encoding="utf-8")
-            data = json.loads(raw)
-        except Exception as exc:
-            _logger.warning("Could not read PMA prompt state: %s", exc)
-            return _default_pma_prompt_state()
-        return data if isinstance(data, dict) else _default_pma_prompt_state()
+def _read_pma_prompt_state_unlocked(path: Path) -> dict[str, Any]:
+    if not path.exists():
+        return _default_pma_prompt_state()
+    try:
+        raw = path.read_text(encoding="utf-8")
+        data = json.loads(raw)
+    except Exception as exc:
+        _logger.warning("Could not read PMA prompt state: %s", exc)
+        return _default_pma_prompt_state()
+    return data if isinstance(data, dict) else _default_pma_prompt_state()
 
 
-def _save_pma_prompt_state(hub_root: Path, state: Mapping[str, Any]) -> None:
-    path = default_pma_prompt_state_path(hub_root)
-    lock_path = _prompt_state_lock_path(path)
-    with file_lock(lock_path):
-        path.parent.mkdir(parents=True, exist_ok=True)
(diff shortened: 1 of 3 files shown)


Git-on-my-level:
Addressed the review feedback in `565e365e`:

- serialized PMA prompt-state read/modify/write under a single file lock so concurrent turns do not clobber stored digests
- fixed notification dropdown indexing so filtered actionable/superseded rows still open the correct backing item

Validation from the commit hook passed, including `pnpm run build`, JS tests, and the full pytest suite (`2884 passed, 1 skipped`).

PR #1023, opened 2026-03-18
Refactor archive flows around explicit intents

### LINKED ISSUE #1019: Refactor CAR archive flows around explicit intents and declarative state specs
## Summary

Archive behavior has accumulated several overlapping concepts and path-specific exceptions that make the code harder to reason about and easier to regress.

Recent work on #1011 / PR #1014 fixed real bugs, but it also exposed a broader design problem:
- archive behavior is driven by overloaded `portable` / `full` profiles
- CAR state handling is stringly-typed and spread across multiple conditional chains
- archive retention policy resolution is not centralized
- snapshot completion is modeled implicitly rather than transactionally

This issue is to clean up the archive model itself.

Note: the `APP_SERVER_EVENT` / `flows.db` payload compaction problem is already tracked separately in #1015 and is out of scope here.

### 1. Archive profile semantics are overloaded

Relevant code:
- `src/codex_autorunner/core/archive.py`
- `src/codex_autorunner/core/hub.py`

`portable` vs `full` now has behavior exceptions such as:
- preserving reset-only state for destructive archive-and-reset flows
- preserving `flows.db` for cleanup-triggered snapshots even when the configured profile is `portable`

That means the effective behavior is no longer really just "portable" or "full". The code is instead expressing different archive intents through boolean exceptions layered on top of one profile flag.

### 2. CAR state path behavior is stringly-typed and duplicated

Relevant code:
- `src/codex_autorunner/core/archive.py`

The same path set has to answer multiple questions:
- is this path dirty?
- should it be archived for this operation?
- where does it land in the snapshot?
- should reset delete it afterward?

Today those rules live in multiple ad hoc conditionals keyed by path strings like `"flows.db"`, `"filebox"`, `"state.sqlite3"`, etc. This is brittle and review-regression-prone.

### 3. Archive retention policy construction is not centralized

Relevant code:
- `src/codex_autorunner/surfaces/cli/commands/cleanup.py`
- `src/codex_autorunner/core/flows/archive_helpers.py`
- `src/codex_autorunner/core/hub.py`

We already had at least one regression where a manual cleanup path read raw config differently from the automatic pruning paths. Operational code should not have multiple ways to assemble archive retention policy.

### 4. Snapshot completion is inferred instead of finalized transactionally

Relevant code:
- `src/codex_autorunner/core/archive.py`
- `src/codex_autorunner/core/archive_retention.py`

The current fix is to ignore snapshot directories without `META.json` during pruning. That avoids pruning in-progress snapshots, but it is still a symptom of a weaker model: a snapshot directory becomes visible before it is fully finalized.

A stronger model would stage into a temp directory and atomically rename into the final snapshot location only after metadata is written.

### A. Replace profile exceptions with explicit archive intents

Instead of treating everything as `portable` or `full` plus exceptions, introduce explicit archive intents such as:
- review snapshot (non-destructive)
- cleanup snapshot (destructive, must preserve reviewable flow state)
- archive-and-reset CAR state (destructive, must preserve everything being reset)

The goal is for call sites to express intent directly and for archive behavior to be derived from that intent, not from layered booleans.

### B. Introduce a declarative CAR state spec table

Represent CAR state paths through a single declarative structure that captures, per path:
- dirty-check strategy
- archive destination
- which archive intents include it
- whether reset removes it
- whether it is runtime-only, review-relevant, or both

This should replace scattered string comparisons and make it much easier to review future archive changes.

### C. Centralize retention policy resolution

Create one helper that resolves worktree/run archive retention policy from parsed config objects. All CLI/API/automatic pruning paths should use the same helper.

### D. Make snapshot finalization transactional

Use a flow like:
1. create temp snapshot directory
2. copy/write contents
3. write `META.json`
4. atomically rename to final snapshot directory
5. only then make it eligible for retention pruning / listing

## Acceptance Criteria

- archive call sites express explicit archive intent instead of relying on profile-plus-exception combinations
- CAR state path rules live in one declarative source of truth
- all retention pruning paths use the same parsed-config policy resolution helper
- snapshot directories are not visible to pruning/listing until finalized
- behavior covered by #1011 / PR #1014 remains preserved by regression tests

## Nice To Have

- add a short architecture note or docs section explaining archive intents and the CAR state spec model
- reduce duplicate test setup by introducing archive-fixture helpers keyed by intent

Git-on-my-level:
## Summary
- replace archive profile exceptions with explicit worktree/archive intents
- centralize CAR state path rules in a declarative spec used for dirty checks, archiving, and reset removal
- centralize worktree/run retention policy resolution and finalize snapshots transactionally before they become visible

## Testing
- .venv/bin/python -m pytest tests/test_archive_worktree.py tests/core/test_archive_retention.py tests/routes/test_archive_routes.py tests/test_hub_supervisor.py tests/test_cli_hub_runs_cleanup.py tests/test_cli_ticket_flow_archive.py -q
- .venv/bin/python -m pytest tests/test_cli_process_diagnostics.py -q
- full pre-commit validation pipeline via git commit hook (3173 passed, 1 skipped)

Closes #1019

### DIFF
--- a/src/codex_autorunner/core/archive.py
+++ b/src/codex_autorunner/core/archive.py
@@ -138,6 +138,17 @@ def _copy_file(src: Path, dest: Path, stats: dict[str, int]) -> None:
     stats["total_bytes"] += dest.stat().st_size
 
 
+def _copy_sqlite_sidecars(src: Path, dest: Path, stats: dict[str, int]) -> None:
+    if not (src.name.endswith(".sqlite3") or src.name.endswith(".db")):
+        return
+    for suffix in SQLITE_SIDE_SUFFIXES:
+        sidecar_src = src.with_name(f"{src.name}{suffix}")
+        if not sidecar_src.exists() or not sidecar_src.is_file():
+            continue
+        sidecar_dest = dest.with_name(f"{dest.name}{suffix}")
+        _copy_file(sidecar_src, sidecar_dest, stats)
+
+
 def _copy_tree(
     src_dir: Path,
     dest_dir: Path,
@@ -216,6 +227,7 @@ def _copy_entry(
 
     if src.is_file():
         _copy_file(src, dest, stats)
+        _copy_sqlite_sidecars(src, dest, stats)
         return True
 
     return False
@@ -587,9 +599,7 @@ def _remove_with_sidecars(path: Path) -> None:
                 sidecar.unlink()
 
 
-def _reset_car_state(worktree_root: Path) -> tuple[str, ...]:
-    from ..bootstrap import seed_repo_files
-
+def _planned_reset_car_state_paths(worktree_root: Path) -> tuple[str, ...]:
     car_root = worktree_root / ".codex-autorunner"
     reset_paths: list[str] = []
     seen: set[str] = set()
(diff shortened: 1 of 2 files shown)


Git-on-my-level:
@codex review

PR #1126, opened 2026-03-25
Refresh hub-managed repos after CAR updates

Git-on-my-level:
## Summary
- refresh hub-managed repo artifacts after every CAR update on both macOS launchd and Linux systemd-user paths
- add a scriptable managed-repo refresh entrypoint that reseeds generated `.codex-autorunner/` files for the hub and each manifest-tracked repo/worktree
- cover the new refresh primitive and wrapper script with focused tests

## Testing
- python -m pytest tests/test_managed_repo_update.py tests/test_ticket_helper_script_sync.py tests/test_system_update_worker.py -q
- bash -n scripts/update-hub-managed-repos.sh scripts/safe-refresh-local-linux-hub.sh scripts/safe-refresh-local-mac-hub.sh
- pre-commit hook: scripts/check.sh (3461 passed, 1 skipped)

### DIFF
--- a/scripts/safe-refresh-local-linux-hub.sh
+++ b/scripts/safe-refresh-local-linux-hub.sh
@@ -140,7 +140,8 @@ _systemd_arg_value() {
   if [[ -z "${service_name}" || -z "${key}" ]]; then
     return 0
   fi
-  systemctl --user cat "${service_name}" 2>/dev/null | "${HELPER_PYTHON}" - "${key}" <<'PY'
+  systemctl --user cat "${service_name}" 2>/dev/null | "${HELPER_PYTHON}" -c "$(
+    cat <<'PY'
 import re
 import sys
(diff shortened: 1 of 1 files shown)


Git-on-my-level:
Addressed the Linux review comment in `120998f0`.

`_systemd_arg_value` now preserves `systemctl --user cat` on stdin and runs the parser via `python -c`, so the unit contents are actually available to the parser instead of being consumed by the heredoc source.

Rechecked with:
- `bash -n scripts/safe-refresh-local-linux-hub.sh scripts/update-hub-managed-repos.sh scripts/safe-refresh-local-mac-hub.sh`
- `python -m pytest tests/test_system_update_worker.py tests/test_managed_repo_update.py -q`
- full `scripts/check.sh` via the repo pre-commit hook

PR #1277, opened 2026-04-07
[codex] Fix Discord slash command ack starvation

Git-on-my-level:
## What changed

This changes the Discord slash-command dispatch path so commands that already defer at dispatch time do not keep running inline on the gateway bypass path when they are known to be long-running.

The fix now queues these commands after the initial interaction ACK:

- `/car new`
- `/car newt`
- `/car session compact`
- `/car session interrupt`

It also adds regression coverage proving a second slash command still gets its defer while a prior deferred command is still executing.

## Why

Some Discord commands were acknowledging the interaction and then continuing to execute as bypass work on the gateway dispatch path. When one of those commands ran long, later slash commands could sit behind it long enough for Discord to reject their callback as a stale interaction.

In production logs this showed up as `10062 Unknown interaction` for `/car newt`, with the command handler itself not being the direct bug.

## User impact

This should make Discord slash commands materially more reliable under contention, especially for session-management commands that can do heavier work like branch resets, thread resets, compaction, or interrupts.

The immediate user-visible symptom addressed here is the flaky `/car newt` timeout, and the same protection now also covers `/car session interrupt` based on the same observed failure pattern.

## Root cause

The initial ACK path and the execution path were split incorrectly:

- CAR deferred the interaction early.
- The command still ran as dispatcher bypass work.
- That blocked later interaction handling on the same gateway worker.
- Later commands then failed their first callback with stale Discord interaction tokens.

## Validation

Local checks run:

- `pytest tests/integrations/discord/test_service_routing.py`
- `pytest tests/integrations/chat/test_command_contract.py`
- repository pre-commit hooks during `git commit`, including full pytest

## Follow-up

This patch is intentionally scoped to the commands with live evidence or matching risk. A better structural follow-up would be to make post-ACK queueing an explicit command-contract property instead of a hardcoded allowlist in the Discord service.


Git-on-my-level:
@codex review

PR #1332, opened 2026-04-09
Harden Discord turn startup failures

Git-on-my-level:
## Summary
- add a watchdog around Discord managed-thread submission so the UI cannot sit on a `working` placeholder forever before a runtime execution even exists
- surface structured log events when Discord timeout/error followups fail instead of swallowing those exceptions
- add regression coverage for startup submission hangs and followup-send failures

## Smoking gun
From the live runtime state in `/Users/dazheng/car-workspace/.codex-autorunner/` after the Discord report on April 10, 2026 local time:

- `discord_state.sqlite3` shows channel `1488827014600331415` switched to PMA thread target `06eb75d0-1b27-4723-b5c4-208482131cfc` at `2026-04-09T17:33:28Z`
- `orchestration.sqlite3` has the corresponding `orch_thread_targets` row for `06eb75d0-1b27-4723-b5c4-208482131cfc`
- that thread target has `runtime_status=idle`, `backend_thread_id=NULL`, and `last_execution_id=NULL`
- there are no `orch_thread_executions` rows at all for that thread target
- Discord had already posted the `working` placeholder in the channel screenshot

That combination means the hang is happening before runtime execution creation, inside the managed-thread submission/startup path. The previous behavior left Discord looking pending forever because the placeholder was posted first and there was no startup watchdog to fail the turn cleanly.

## Testing
- `.venv/bin/pytest -q tests/integrations/discord/test_message_turns.py -k 'submission_timeout or begin_failure'`
- `.venv/bin/pytest -q tests/integrations/discord/test_reliability.py -k 'followup_failure_is_logged'`
- full pre-commit suite on commit


Git-on-my-level:
Addressed the review feedback from [discussion_r3061721823](https://github.com/Git-on-my-level/codex-autorunner/pull/1332#discussion_r3061721823) in `f8d75124`.

The Discord startup watchdog still fails fast, but `_ThreadExecutionLifecycle.start_execution` now catches `asyncio.CancelledError`, records the already-created execution as `error` with the configured public failure detail, and then re-raises the cancellation. That prevents the timeout path from leaving an orphaned `running` execution behind if startup is cancelled after `send_message(...)` has created it.

I also added a regression in `tests/core/orchestration/test_service.py` that forces `send_message(...)` to time out during `start_turn(...)` and asserts the execution is persisted as failed and no running execution remains.

PR #1411, opened 2026-04-14
Reduce orchestration CPU hot paths

Git-on-my-level:


Git-on-my-level:
Addressed the current review feedback on `fix/cpu-usage` in commit `8f51fd87`.

What changed:
- Compaction precheck semantics are preserved: baseline hot rows are counted without `:compaction-summary`, so `max-1 kept rows + summary + one new event` no longer triggers an unnecessary second compaction pass.
- The event-type -> hot-family mapping is centralized in `execution_history.py` and reused by diagnostics, maintenance, and turn-timeline.
- Checkpoint recovery now seeds notice state by projection order (`event_id ASC`) instead of timestamp order, and output-memory recovery uses the same ordering for consistency.

Regression coverage added:
- A compaction test that proves summary rows do not force re-compaction at the threshold boundary.
- A turn-timeline test that seeds out-of-order timestamps and verifies the latest notice memory is rebuilt from projection order.

Validation:
- `rtk .venv/bin/python -m pytest tests/core/orchestration/test_execution_history.py tests/core/orchestration/test_execution_history_maintenance.py tests/core/orchestration/test_turn_timeline.py -q`
- `rtk .venv/bin/python -m pytest tests/core/orchestration -q`

Git-on-my-level:
Re: **Duplicated event-type-to-family mapping across two files** — already addressed in [commit `99d0d5d1`](https://github.com/Git-on-my-level/codex-autorunner/commit/99d0d5d1). Both `turn_timeline.py` and `execution_history_diagnostics.py` now import the canonical `timeline_hot_family_for_event_type` from `execution_history.py` instead of maintaining separate copies.

Git-on-my-level:
Re: **Seed notice state by projection order, not timestamp** — already addressed. `_seed_notice_memory` at `turn_timeline.py:141` uses `ORDER BY event_id ASC`, not timestamp. The only remaining `ORDER BY timestamp ASC, event_id ASC` is in `list_turn_timeline` (line 882), which is a display-only listing API — not a state recovery path — so timestamp-first ordering is correct there.

PR #1423, opened 2026-04-14
Fix Discord flow archive store-open error classification

Git-on-my-level:
## Summary
- Correct error classification in Discord `/flow` archive button handling when `FlowStore` fails to open with a SQLite-layer error.
- Keep transient error mapping and user messaging deterministic: store-open failures now consistently log `discord.flow.store_open_failed` and return "Unable to access flow database...".
- Add regression coverage for SQLite open failures raised by `_open_flow_store`.

## Root Cause
The archive callback previously classified exceptions by broad type *after* `to_thread(...)` returned. When `_open_flow_store(...)` raised `sqlite3.OperationalError`, it was treated as a query failure (`discord.flow.query_failed`) because the type matched `sqlite3.Error`, even though the failure happened before query execution.

## Fix
- Move open/query classification into `_resolve_archive_target()` itself:
  - `_open_flow_store(...)` errors -> `discord.flow.store_open_failed` + store-access user message.
  - `_resolve_flow_run_by_id(...)` errors -> `discord.flow.query_failed` + query user message.
- Re-raise only `DiscordTransientError` from the threaded call path.

## Validation
- Focused: `.venv/bin/pytest tests/integrations/discord/test_flow_archive.py -q`
- Repo gate via commit hooks (passed): formatting, ruff, mypy, frontend build/tests, full pytest (`5135 passed`).

### DIFF
--- a/src/codex_autorunner/integrations/discord/flow_commands.py
+++ b/src/codex_autorunner/integrations/discord/flow_commands.py
@@ -2176,36 +2176,50 @@ async def handle_flow_button(
             )
             return
 
-        try:
+        def _resolve_archive_target() -> Optional[FlowRunRecord]:
             store = service._open_flow_store(workspace_root)
-        except (sqlite3.Error, OSError, RuntimeError, ConfigError) as exc:
-            log_event(
-                service._logger,
-                logging.ERROR,
-                "discord.flow.store_open_failed",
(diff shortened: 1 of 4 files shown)


Git-on-my-level:
Superseded by #1424 (same fix, clean single-purpose diff with only the Discord archive classification change).

PR #1437, opened 2026-04-14
[codex] Preserve resumable backend threads across runtime restarts

Git-on-my-level:
## Summary

This fixes a session-recovery bug where Discord/CAR follow-up messages could immediately show:

`Notice: the previous live session was unavailable, so I started a new session.`

instead of reusing the existing durable backend thread.

## Root Cause

The orchestration layer treated a changed `backend_runtime_instance_id` as proof that the prior backend conversation was dead.

In practice, for Codex-backed threads that runtime instance id is tied to the local app-server process, not the durable backend thread identity. After a repo/app-server restart, the runtime instance id changes even when the underlying conversation id is still resumable.

Because of that, `start_execution()` cleared the stored `backend_thread_id` before attempting `resume_conversation()`, which forced a fresh backend session on the next follow-up turn.

## What Changed

- Stop preemptively clearing the backend thread binding when only the runtime instance id changed.
- Log the condition as a stale binding hint, but attempt `resume_conversation()` first.
- Keep the existing fallback behavior: if resume/start still proves the thread is gone, orchestration will start a fresh backend conversation and emit the recovery notice.
- Update the orchestration test to assert that a runtime-instance change alone no longer forces a fresh backend session.

## Impact

- Follow-up turns can survive harmless app-server/runtime restarts as long as the durable backend conversation is still resumable.
- The recovery notice should now appear only when the backend thread is actually unavailable, not merely when the local runtime process changed.

## Validation

- `pytest tests/core/orchestration/test_service.py -k "backend_runtime_instance or runtime_binding_restart or recoverable_backend_error or fresh_when_backend_runtime"`
- `pytest tests/integrations/discord/test_message_turns_backend_binding.py tests/integrations/chat/test_managed_thread_turns.py -k "backend_binding or resolve_managed_thread_target"`
- Full pre-commit hook via `git commit`, including strict typecheck, frontend build/tests, and repo fast-test suite (`5177 passed` on the successful retry)

## Notes

I also inspected local CAR runtime artifacts while debugging:

- `.codex-autorunner/state.sqlite3` had no reusable terminal session rows.
- The local repo `orchestration.sqlite3` was not authoritative for this Discord path; the managed-thread path is using the hub control plane.
- The problematic behavior matched the orchestration stale-runtime branch rather than a rendering-only issue.


Git-on-my-level:
@cursor sync with origin main and fix CI

PR #1486, opened 2026-04-17
[codex] Repair SCM follow-up when PR thread is archived

Git-on-my-level:
## Summary
This repairs SCM follow-up delivery when a PR binding still points at an archived or missing managed thread.

## Root Cause
SCM automation kept the original `thread_target_id` from the PR binding even after that managed thread had been archived. `enqueue_managed_turn` then retried delivery against the same inactive thread, which raised `ManagedThreadNotActiveError` on every attempt. After the retry budget was exhausted, the automation layer escalated with the delivery failure message instead of recovering the PR follow-up flow.

## What Changed
- detect stale SCM-bound managed threads during `enqueue_managed_turn` execution
- create a replacement managed thread in the PR workspace and rebind the PR to it
- enqueue a bootstrap prompt for the fresh session with the PR, branch, trigger, review-comment focus, and `gh pr` commands to inspect/update the PR
- prefer the executor response thread id when sending the review-batch notice after a successful enqueue
- keep head/base branch details in SCM tracking metadata so the replacement session has the right PR context
- add regression coverage for archived-thread rebinding and the notice target selection

## Impact
PR review feedback and CI follow-ups now recover by starting a fresh managed session instead of failing three times against an archived thread and escalating.

## Validation
- `.venv/bin/pytest tests/core/test_publish_executor.py tests/core/test_scm_automation_service.py`
- repo pre-commit validation lane (`black`, `ruff`, strict `mypy`, fast-test budget verification, dead-code check)
- full non-integration pytest lane via pre-commit: `5746 passed, 9 xfailed`

### DIFF
--- a/src/codex_autorunner/core/pma_thread_store.py
+++ b/src/codex_autorunner/core/pma_thread_store.py
@@ -1290,6 +1290,40 @@ class PmaThreadStore:
             return None
         return _execution_row_to_record(row)
 
+    def get_turn_by_client_turn_id_any_thread(
+        self, client_turn_id: str
+    ) -> Optional[dict[str, Any]]:
+        """Return the best matching execution for this client id across all threads.
+
+        Publish dedupe keys do not vary when a PR binding is repointed to a new
+        managed thread; without a global lookup, a retried enqueue could miss a
+        turn created on the replacement thread and enqueue a duplicate.
+        """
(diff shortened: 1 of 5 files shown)


Git-on-my-level:
@cursor can you address the review comment from codex?

Issue #1466, opened 2026-04-16
Modularize the ticket flow live output UI and add a shared frontend harness


Git-on-my-level:
## Problem
`src/codex_autorunner/static_src/tickets.ts` currently carries too much responsibility. The live agent output panel logic is mixed in with ticket list behavior, dispatch history, panel controls, and flow state.

That makes small behavior changes harder to reason about and increases the risk of UI state bugs.

## Why this matters
- the live output panel has already accumulated multiple presentation modes and state transitions
- future agents have to rediscover the same local architecture each time
- frontend integration tests currently re-bootstrap JSDOM/DOM globals ad hoc instead of reusing a shared harness

## Proposed scope
- extract the ticket flow live output panel into a dedicated module with a clear state model
- keep a small explicit API for collapsed/expanded state, status updates, plain-text output, and event rendering
- add a reusable frontend/JSDOM test harness for generated static modules so tests stop hand-rolling the same setup
- move or add focused tests around the extracted module

## Acceptance criteria
- live output panel behavior is implemented outside the main `tickets.ts` monolith
- the extracted module has focused tests covering collapsed/expanded behavior and stream rendering
- at least one shared frontend test helper exists for DOM/global bootstrap used by this surface

PR #1491, opened 2026-04-18
[codex] Fix validation lane routing and live output panel UX

### LINKED ISSUE #1465: Keep web UI changes out of aggregate validation when possible
## Problem
Small web UI changes can escalate into the full `aggregate` validation lane when a related test is classified as `core` instead of `web-ui`.

A concrete example is `tests/test_ticket_flow_ui_integration.py`: when changed alongside `src/codex_autorunner/static_src/*`, the mixed lane selection falls back to `aggregate`, which triggers the full hook suite for what is otherwise a scoped frontend change.

## Why this matters
- slows down iteration for UI-only work
- makes agent/operator feedback noisier than necessary
- obscures the actual reason the full lane was selected

## Proposed scope
- classify ticket-flow web UI integration tests under the `web-ui` lane, or move them under a path that already maps there
- review other root-level frontend-oriented tests that currently default to `core`
- improve `scripts/check.sh` / lane reporting so the selected lane explains which paths caused escalation

## Acceptance criteria
- a UI-only change in `static_src` plus its matching frontend integration test does not automatically fall back to `aggregate`
- lane selection output clearly reports the reason and the paths that triggered it
- behavior remains conservative for genuinely shared-risk or multi-surface diffs

Git-on-my-level:
## Summary
- keep ticket-flow UI integration and other clearly frontend root tests in the `web-ui` validation lane
- include lane-path reporting so aggregate escalation explains exactly which files triggered it
- simplify the repo agent live-output panel to a single header-toggle interaction with a quieter streaming highlight
- related live-output modularization follow-up tracked in `#1466`

## Validation
- `pnpm run build`
- `pnpm test:markdown`
- `.venv/bin/python -m pytest tests/core/test_validation_lanes.py tests/test_select_validation_lane_script.py tests/test_select_ci_validation_jobs_script.py tests/test_ticket_flow_ui_integration.py`
- aggregate pre-commit hook (`scripts/check.sh`) during `git commit`

Closes #1465
Related: #1466

### DIFF
--- a/scripts/check.sh
+++ b/scripts/check.sh
@@ -84,11 +84,21 @@ need_cmd make
 if [[ -z "$LANE" ]]; then
   STAGED_FILES="$(git diff --cached --name-only --diff-filter=ACM 2>/dev/null || true)"
   if [[ -n "$STAGED_FILES" ]]; then
-    LANE="$(echo "$STAGED_FILES" | "$PYTHON_BIN" scripts/select_validation_lane.py --format json | "$PYTHON_BIN" -c 'import json,sys; print(json.load(sys.stdin)["lane"])')"
+    SELECTION_TEXT="$(printf '%s\n' "$STAGED_FILES" | "$PYTHON_BIN" scripts/select_validation_lane.py)"
+    LANE="$(
+      printf '%s\n' "$STAGED_FILES" \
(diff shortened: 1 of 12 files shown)


Git-on-my-level:
@cursor please fix CI

PR #1506, opened 2026-04-18
[codex] Add durable managed-thread final delivery and replay

### LINKED ISSUE #1498: Managed thread delivery pipeline silently drops messages on transient failures
## Problem

When a managed thread turn finalizes successfully but the subsequent delivery callback fails (or is cancelled), the agent response is permanently lost with no log trace. The user sees no response in Discord/Telegram until a hub restart recovers the pending state.

### Observed incident (2026-04-18)

- Discord car-1 channel, codex agent completed a turn at 15:16:40 local
- `turn_finalized` logged successfully (status=ok, 352 chars)
- The delivery chain (`managed_thread_submission` → `managed_thread_finalized` → `delivery_started`) never fired
- No error was logged for the missing delivery
- Response appeared in Discord ~2.75 hours later after a hub restart
- Hub control plane had 3 ReadTimeout/ReadError events in the preceding 3 hours (15:05, 15:15, 16:38)

### Root causes in `managed_thread_turns.py`

**1. Wrong exception types in best-effort hub call (line 2199)**

```python
except (HubControlPlaneError, RuntimeError, OSError):
```

`httpx.ReadTimeout` inherits from `Exception → httpx.TimeoutException`, not from any of these three. A control plane timeout propagates uncaught when it should be swallowed (this is explicitly best-effort).

**2. CancelledError re-raised without logging (line 1206-1207)**

```python
if isinstance(exc, asyncio.CancelledError):
    raise
```

If the queue worker task is cancelled during finalization (e.g., gateway reconnect cleanup, indicator teardown), the delivery is lost and zero log output is produced. This is the most likely explanation for the observed incident — no error log exists for the missing delivery.

**3. deliver_result drops messages permanently (line 1241-1253)**

```python
try:
    await deliver_result(finalized)
except (RuntimeError, OSError, ValueError, TypeError, ConnectionError):
    logger.exception(...)
```

Even when the error IS caught and logged, the finalization result is discarded. There is no retry, no dead-letter queue, no persistence. For chat surfaces where a user is waiting for a response, the message is gone until a restart.

**4. run_with_indicator can mask the exception source (line 1197)**

The indicator wrapper runs finalization as a sub-callback. If the indicator itself raises or cancels subtasks, the original exception context is lost.

### A. Every exception path produces a log entry before any re-raise or return

No code path should silently swallow an exception. This includes:

- `asyncio.CancelledError` — log with full context (thread ID, execution ID, channel ID) before re-raising
- Any exception propagating out of `run_with_indicator` — log the indicator state and the original exception
- Hub control plane calls — catch `Exception` instead of narrow types for best-effort calls

### B. Undelivered finalization results are persisted to a replay queue

When `deliver_result` fails, write the finalization result to a durable replay queue before giving up:

- Append to a JSONL file (e.g., `.codex-autorunner/delivery_dead_letter.jsonl`)
- Include: managed_thread_id, execution_id, channel_id, surface, assistant_text, timestamp, error
- On hub startup, scan the dead-letter queue and attempt redelivery for entries younger than a configurable TTL (e.g., 1 hour)
- After successful redelivery, remove the entry
- This is the minimum-viable fix for the "response stuck until restart" class of bug

### C. Hub control plane best-effort calls catch Exception

For any call annotated as best-effort (activity recording, cold trace persistence), catch `Exception` instead of a narrow type list. The current narrow catches are a maintenance hazard — any new exception type from httpx/aiohttp/etc. will propagate unexpectedly.

### D. Delivery pipeline has at-least-once semantics

The delivery pipeline should guarantee at-least-once delivery for final messages:

- Write the finalization result to the dead-letter queue *before* attempting delivery
- On successful delivery, remove from the queue
- On failure, leave in the queue for retry on next startup or periodic sweep
- Idempotency key (execution_id) prevents double-delivery on retry

### E. Structured error context in all exception handlers

Every `except` block in the delivery pipeline should include:

- `managed_thread_id`
- `execution_id` / `managed_turn_id`
- `channel_id` / surface
- `backend_thread_id`
- The full exception chain (`exc_info=True`)

This makes it possible to correlate errors across the hub log, discord bridge log, and any dead-letter queue entries.

## Scope note

This issue describes the ideal end state. Individual PRs can tackle subsets:

1. Quick win: catch `Exception` for best-effort calls + log before CancelledError re-raise (trivial change, high diagnostic ROI)
2. Medium: dead-letter queue for undelivered results
3. Larger: at-least-once delivery semantics with startup replay

Git-on-my-level:
## Summary
This PR makes managed-thread final delivery durable across Discord and Telegram by introducing a control-plane delivery ledger, replay/recovery workers, and adapter wiring for queued and direct finalization paths.

It also includes a final-review fix for a delivery-engine starvation bug: active `claimed`/`delivering` records no longer appear in the due queue, and `claim_next_delivery()` now performs explicit expired-claim recovery before selecting the next ready delivery.

## What Changed
- add the durable managed-thread delivery contract, SQLite ledger, migration, claim/retry state machine, and recovery logic
- wire Discord and Telegram final delivery through durable intents, adapters, and background delivery workers
- expand observability and regression coverage for delivery replay, recovery, worker behavior, and surface adapter integration
- tighten the due-delivery query and add regressions for active-claim starvation and exhausted-claim abandonment

## Why
Managed-thread final delivery was still relying on surface-local callbacks and transport timing, which made replay/recovery fragile after cancellations, crashes, or lease expiry. The new ledger makes delivery intent durable before transport IO and gives both chat surfaces a shared recovery path.

The review fix closes a correctness hole where an in-flight claimed record could block later pending work on the same adapter.

## Impact
- finalized managed-thread responses can be replayed and recovered durably
- Discord and Telegram now share the same delivery lifecycle and retry semantics
- expired claims are recovered explicitly without starving later due deliveries

closes https://github.com/Git-on-my-level/codex-autorunner/issues/1498

## Validation
- repo validation lane during commit: `black`, `ruff`, import-boundary checks, hub contract checks, strict `mypy`, and full `pytest` (`5900 passed, 9 xfailed`)
- focused delivery suites: `./.venv/bin/pytest tests/core/orchestration/test_managed_thread_delivery.py tests/core/orchestration/test_managed_thread_delivery_ledger.py tests/core/orchestration/test_managed_thread_delivery_recovery.py -q`
- focused integration suites: `./.venv/bin/pytest tests/integrations/chat/test_hermes_official_completion.py tests/integrations/chat/test_managed_thread_delivery_intent.py tests/integrations/chat/test_managed_thread_delivery_worker.py tests/integrations/chat/test_managed_thread_turns.py tests/integrations/chat/test_managed_turn_runner.py tests/integrations/discord/test_discord_managed_thread_delivery_adapter.py tests/integrations/telegram/test_telegram_managed_thread_delivery_adapter.py tests/test_housekeeping.py tests/test_telegram_session_recovery_notice.py -q`


Git-on-my-level:
@cursor resolve merge conflicts

Git-on-my-level:
@cursor address the new codex review comments and failing CI

Git-on-my-level:
@cursor review

PR #1524, opened 2026-04-20
[codex] Improve Discord PMA toggle session UX

Git-on-my-level:
## What changed
- make Discord `/pma on` and `/pma off` create a fresh managed-thread session and return the same detailed status block used for fresh session startup
- preserve the compact-seed cleanup behavior even when a PMA toggle fails mid-transition
- extract the PMA mode-switch flow into a dedicated helper module so the Discord PMA command router stays within hotspot budget caps
- extend the chat-surface lab runner with PMA slash-command interactions and add a deterministic `pma_mode_toggle` scenario
- update Discord PMA command tests to assert the new session-report UX

## Why
PMA mode toggles were only flipping binding state and replying with a one-line confirmation, which left the user without the new thread id or session metadata and made PMA mode transitions feel inconsistent with `/car new`. The desired behavior is to treat PMA mode entry and exit as fresh session boundaries and show the same new-session report.

## User impact
- entering PMA mode now shows the fresh PMA session report immediately
- leaving PMA mode now starts a fresh repo session when a saved workspace binding exists and shows the repo session report
- if no saved workspace binding exists, `/pma off` reports that the channel is now unbound instead of pretending a repo session was restored
- the chat lab can now reproduce this flow directly

## Root cause
The PMA toggle handlers in Discord mutated `pma_enabled` state but did not call the managed-thread reset path used by the existing fresh-session workflow.

## Validation
- `.venv/bin/python -m pytest -q tests/integrations/discord/test_pma_commands.py`
- `.venv/bin/python -m pytest -q tests/chat_surface_lab/test_scenario_runner.py -k pma_mode_toggle tests/chat_surface_lab/test_scenario_corpus.py`
- commit hook validation passed, including strict mypy, full pytest, frontend build/tests, and chat-surface latency checks

## Review note
A mini subagent review found one regression risk around stale compact-seed cleanup on PMA toggle failures. That was addressed before this PR was opened.


Git-on-my-level:
@cursor address the codex review comment

Git-on-my-level:
@cursor resolve merge conflicts on pr

PR #1525, opened 2026-04-20
[codex] Fix web repo settings modal wiring

Git-on-my-level:
## Summary
- factor web and repo settings update-target loading/update actions through one shared frontend module
- wire repo settings to hydrate and save autorunner model/effort/approval/sandbox state from `/api/session/settings`
- add frontend and route coverage for repo settings hydration, save/reload, and failure handling

## Root cause
The repo settings modal had backend routes for session settings, but the frontend never loaded them, so the select controls rendered blank and save behavior was incomplete. The web update surface also had duplicated client logic instead of reusing a shared path.

## Validation
- git commit hook lane `web-ui`
- repo-wide `mypy`
- repo-wide `pytest` (`7517 passed`)
- `pnpm run build`
- `pnpm test:markdown`

### DIFF
--- a/src/codex_autorunner/static/assets.json
+++ b/src/codex_autorunner/static/assets.json
@@ -152,6 +152,9 @@
       "path": "generated/streamUtils.js"
     },
     {
+      "path": "generated/systemUpdateUi.js"
+    },
+    {
       "path": "generated/tabs.js"
     },
     {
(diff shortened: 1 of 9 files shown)


Git-on-my-level:
@cursor address the codex review comment

PR #1533, opened 2026-04-20
[codex] flush discord outbox after durable delivery

Git-on-my-level:
## What changed
Managed-thread Discord durable delivery now carries the originating `workspace_root` in the delivery intent metadata and flushes FileBox outbox artifacts after a successful durable terminal delivery.

## Why it changed
Agents can place reply artifacts in `.codex-autorunner/filebox/outbox/` instead of `outbox/pending/`. Normal Discord turn cleanup already flushes those files, but managed-thread durable delivery bypassed that cleanup path when `send_final_message` was false.

## Root cause
The durable managed-thread adapter delivered the final Discord message without performing the post-turn FileBox flush, so outbox-root artifacts could be stranded even though the text response was delivered.

## Impact
Discord managed-thread runs now send outbox-root artifacts after durable terminal delivery, including replayed durable deliveries. The regression test coverage now asserts that the durable adapter triggers the outbox flush with the correct workspace path.

## Validation
- Commit hook lane `chat-apps`
- `python -m pytest -q tests/integrations/discord/test_discord_managed_thread_delivery_adapter.py`
- `python -m pytest -q tests/integrations/discord/test_message_turns_message_flow.py`
- `python -m pytest -q tests/integrations/discord/test_service_routing.py -k "managed_thread_delivery or token_usage_footer"`
- Repo hook checks: black, ruff, import/contracts, mypy, full pytest (`7556 passed`), chat-surface deterministic checks

### DIFF
--- a/src/codex_autorunner/integrations/chat/managed_thread_delivery_support.py
+++ b/src/codex_autorunner/integrations/chat/managed_thread_delivery_support.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import asyncio
+import logging
 from dataclasses import dataclass, field
 from typing import Any, Awaitable, Callable, Mapping, Optional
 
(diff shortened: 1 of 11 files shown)


Git-on-my-level:
@cursor please address failing CI systematically. If there are other flakes fix them as well

PR #1541, opened 2026-04-21
[codex] Harden Hermes PMA timeout recovery

Git-on-my-level:
## What changed
- add Hermes session-store recovery for stalled ACP turns and thread it through the ACP client, Hermes supervisor/harness, and runtime-thread stall handling
- reconcile cancelled Discord PMA startup tasks so stuck progress state is cleaned up visibly instead of being left hanging
- harden Telegram and Discord chat-surface harness draining/shutdown behavior so timeout-path tests are deterministic
- extract Discord reusable-progress runtime state out of `progress_leases.py` to get back under the hotspot seam budget

## Why
We were seeing two recurring PMA failure modes:
1. Hermes could finish internally and persist the assistant reply in its session store, but CAR never received the ACP terminal boundary and timed the turn out.
2. Discord PMA startup/background-task cancellation could leave execution/progress state in a bad surface state and make reproductions flaky.

The Telegram harness also had a pre-existing `CancelledError` drain bug, and the Discord progress lease seam was over the repo budget cap.

## Root cause
- Hermes recovery baseline snapshots were being taken after `start_prompt`, which could race with same-turn persisted output and suppress recovery.
- Runtime-thread stall handling had no recovery hook before timing out.
- Discord background-task reconciliation handled `CancelledError`, but shutdown/startup failure paths still needed deterministic cleanup coverage.
- Telegram drain helpers treated intentionally cancelled spawned tasks as hard harness failures.
- `progress_leases.py` had drifted past its enforced seam budget.

## Impact
- stalled Hermes PMA turns can recover from the real Hermes session store instead of surfacing a false timeout when the terminal ACP event is missing
- Discord PMA startup failures and timeout-path regressions are reproducible and covered at ACP, supervisor, orchestration, and chat-surface layers
- Telegram timeout-path harnesses no longer fail on expected task cancellation
- Discord progress-lease code is back under the hotspot cap without changing external call sites

## Validation
- full repo hook during commit passed, including:
  - black
  - ruff
  - import-boundary checks
  - strict mypy
  - full pytest: `7618 passed`
  - frontend build/tests
  - chat-surface lab deterministic checks and latency budget suite
- focused regressions also exercised:
  - `tests/agents/acp/test_client.py`
  - `tests/agents/hermes/test_hermes_supervisor_official_prompt_hang.py`
  - `tests/core/orchestration/test_runtime_threads.py`
  - `tests/chat_surface_integration/test_hermes_pma_official_timeout.py`
  - `tests/test_discord_message_turn_startup_pending.py`
  - `tests/chat_surface_integration/test_hermes_pma_ux_regressions.py`
  - `tests/chat_surface_integration/test_hermes_pma_surfaces.py`
  - `tests/test_hotspot_budgets.py`

### DIFF
--- a/src/codex_autorunner/agents/acp/client.py
+++ b/src/codex_autorunner/agents/acp/client.py
@@ -1084,6 +1084,47 @@ class ACPClient:
             completion_source=completion_source,
         )
 
+    async def recover_prompt_completion(
+        self,
+        turn_id: str,
+        *,
+        final_output: str,
+        recovery_source: str = "session_store",
+    ) -> bool:
+        state = self._prompts.get(turn_id)
+        if state is None or state.closed:
+            return False
+        terminal_event = normalize_notification(
+            {
+                "method": "prompt/completed",
+                "params": {
+                    "sessionId": state.session_id,
+                    "turnId": state.turn_id,
+                    "status": "completed",
+                    "finalOutput": final_output,
+                    "recoveredFrom": recovery_source,
+                },
+            }
+        )
+        if not isinstance(terminal_event, ACPTurnTerminalEvent):
+            raise ACPProtocolError(
+                "Recovered ACP prompt completion did not normalize terminally"
+            )
+        await self._record_prompt_terminal_event(
+            state,
+            terminal_event,
+            completion_source=recovery_source,
+        )
+        self._log_trace_event(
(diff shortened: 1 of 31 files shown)


Git-on-my-level:
@cursor address codex review comment and merge conflict on PR

PR #1551, opened 2026-04-22
Clarify live commentary progress

Git-on-my-level:
## Summary
- make shared chat progress rendering label commentary as explicitly in-progress rather than looking like a normal assistant reply
- stop Discord PMA from suppressing the visible final reply when durable delivery is only pending or retry-scheduled
- add regression coverage for the pending durable-delivery case and update the chat-surface lab scenario to expect the new direct-fallback behavior

## Why
Two issues were compounding the confusing UX:
1. Hermes commentary looked too much like a finished answer while the turn was still running.
2. Discord PMA treated pending durable delivery as if final visibility was already handled, which could delay the user-visible final reply until a later retry.

The second issue is the root-cause fix here: Discord now matches Telegram semantics and only suppresses its visible final reply after a confirmed durable delivery.

## Impact
- Discord managed-thread progress is clearer while a turn is still running
- Discord PMA no longer delays the visible final reply just because durable delivery is pending
- Telegram and Discord now align on the durable-delivery suppression rule

## Checks
- `pytest tests/discord_message_turns_support.py -q -k pending_durable_delivery`
- `pytest tests/integrations/discord/test_message_turns_transient_progress.py -q -k pending_durable_delivery`
- `pytest tests/chat_surface_integration/test_hermes_pma_surface_parity.py -q -k official_completion`
- `pytest tests/chat_surface_integration/test_hermes_pma_ux_regressions.py -q -k 'fast_visible_feedback_and_reuse_progress_anchor or make_busy_thread_queue_visible_before_recovery'`
- `pytest tests/chat_surface_lab/test_scenario_runner.py -q -k final_delivery_retry_after_visibility`
- `pytest tests/test_hotspot_budgets.py -q -k hotspot_file_budgets`
- aggregate commit lane passed, including strict mypy, full pytest, frontend checks, and chat-surface lab/latency checks

### DIFF
--- a/src/codex_autorunner/integrations/chat/progress_primitives.py
+++ b/src/codex_autorunner/integrations/chat/progress_primitives.py
@@ -410,12 +410,23 @@ def render_progress_text(
             block = [f"🧠 {action.text}"]
             if blocks:
                 block.insert(0, "")
-        elif action.label in {"output", "commentary"}:
+        elif action.label == "output":
             output_lines = action.text.split("\n")
             if not output_lines:
                 block = [action.text]
             else:
                 block = output_lines
+            if blocks:
+                block.insert(0, "")
+        elif action.label == "commentary":
+            commentary_lines = action.text.split("\n")
+            if not commentary_lines:
+                commentary_lines = [action.text]
+            block = [
+                "Interim note from agent while this turn is still running:",
+                *commentary_lines,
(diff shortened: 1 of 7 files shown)


Git-on-my-level:
@codex review

Git-on-my-level:
@cursor review

PR #1552, opened 2026-04-22
[codex] Fix PMA duplicate delivery and Hermes progress UX

Git-on-my-level:
## What changed
- preserve the durable managed-thread `delivery_id` through finalization so surfaces know which pending terminal delivery record belongs to the current turn
- abandon pending durable deliveries after a direct visible final reply succeeds on Discord and Telegram, preventing the background delivery worker from replaying the same finalized turn later
- remove the commentary decorator wrapper from the shared progress renderer and let Hermes commentary use the normal working edit streaming flow
- dedupe commentary that matches already-streamed live output so the progress card does not show the same text twice
- add regression coverage for the shared runner, Discord final delivery, Telegram PMA delivery, and commentary rendering behavior

## Why
Hermes PMA turns could show two bad behaviors after deployment:
- duplicate replies in Discord and Telegram when a direct final reply succeeded but a pending durable delivery record was still retried later
- apparent off-by-one replies where a past finalized turn arrived after a newer user message, because the replay path stayed alive

The interim commentary wrapper also made Hermes progress noisier than Codex/OpenCode and could visually duplicate text already present in the working stream.

## Root cause
The shared managed-turn flow only carried boolean durable-delivery state, not the actual durable `delivery_id`. That meant Discord and Telegram could not retire a pending durable terminal delivery after a direct visible reply succeeded, so the worker could replay the same turn later. Separately, commentary was rendered as a second decorated block instead of flowing through the existing live working-edit UX.

## Impact
- fixes duplicate final delivery and stale-turn replay on Discord and Telegram PMA turns
- keeps direct final reply fallback behavior fast while closing the replay window
- makes Hermes commentary match the existing working progress UX instead of adding extra wrapper text

## Validation
- aggregate commit lane passed locally
- `.venv/bin/pytest -q tests/integrations/chat/test_managed_turn_runner.py`
- `.venv/bin/pytest -q tests/integrations/discord/test_message_turns_transient_progress.py`
- `.venv/bin/pytest -q tests/telegram_pma_managed_thread_support.py`
- `.venv/bin/pytest -q tests/test_managed_thread_progress.py -k commentary`
- `.venv/bin/pytest -q tests/test_telegram_progress_stream.py -k commentary`
- `.venv/bin/pytest -q tests/chat_surface_lab/test_scenario_runner.py -k durable_delivery_is_outboxed`
- `.venv/bin/python -m py_compile src/codex_autorunner/integrations/chat/managed_thread_progress.py src/codex_autorunner/integrations/chat/managed_thread_turns.py src/codex_autorunner/integrations/chat/managed_turn_runner.py src/codex_autorunner/integrations/chat/progress_primitives.py src/codex_autorunner/integrations/discord/message_turns.py src/codex_autorunner/integrations/telegram/handlers/commands/execution.py src/codex_autorunner/integrations/telegram/handlers/commands_runtime.py tests/integrations/chat/test_managed_turn_runner.py tests/integrations/discord/test_message_turns_transient_progress.py tests/telegram_pma_managed_thread_support.py tests/test_managed_thread_progress.py tests/test_telegram_progress_stream.py`

### DIFF
--- a/src/codex_autorunner/integrations/chat/managed_thread_progress.py
+++ b/src/codex_autorunner/integrations/chat/managed_thread_progress.py
@@ -14,7 +14,12 @@ from ...core.ports.run_event import (
     RunNotice,
     ToolCall,
 )
-from .progress_primitives import ProgressAction, TurnProgressTracker
+from .progress_primitives import (
+    ProgressAction,
+    TurnProgressTracker,
+    _normalize_output_text,
+    _output_matches_final_message,
+)
 
 
 @dataclass
@@ -123,6 +128,18 @@ def apply_run_event_to_progress_tracker(
             if already_streamed:
                 return ProgressTrackerEventOutcome(changed=False)
             if not notice:
+                return ProgressTrackerEventOutcome(changed=False)
+            latest_output = _normalize_output_text(tracker.latest_output_text()).strip()
+            notice_text = _normalize_output_text(notice).strip()
+            if (
+                latest_output
+                and notice_text
+                and (
+                    latest_output == notice_text
+                    or _output_matches_final_message(latest_output, notice_text)
+                    or _output_matches_final_message(notice_text, latest_output)
+                )
+            ):
                 return ProgressTrackerEventOutcome(changed=False)
             tracker.note_commentary(notice)
             return ProgressTrackerEventOutcome(changed=True, force=True)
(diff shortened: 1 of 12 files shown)


Git-on-my-level:
@cursor address failing CI

Issue #1553, opened 2026-04-22
Follow up PMA delivery and progress footguns after duplicate-delivery fix


Git-on-my-level:
## Context
PR #1552 fixes the immediate Hermes PMA regression where Discord and Telegram could duplicate finalized replies or deliver an older finalized turn after a newer user message. While implementing that fix, a few remaining footguns became clear.

This issue tracks the follow-up cleanup work so the current patch does not become the long-term architecture.

### 1. Replace post-send abandonment with a first-class direct-delivery ledger state
Current behavior:
- the surface posts a direct visible final reply
- then the code abandons the pending durable delivery record to stop worker replay

That closes the common replay path, but there is still a narrow race if a worker claims or delivers the durable record before the abandon lands.

Prescription:
- add a first-class ledger transition for something like `DIRECT_SURFACE_DELIVERED` or `SUPPRESSED_BY_DIRECT_DELIVERY`
- let the direct-send path claim or atomically transition the durable record before or during surface delivery, rather than abandoning it afterward
- make worker claim/recovery logic treat that state as terminal and non-deliverable
- update the replay/ledger tests to cover the race explicitly

Likely touch points:
- `src/codex_autorunner/integrations/chat/managed_turn_runner.py`
- `src/codex_autorunner/core/orchestration/managed_thread_delivery_ledger.py`
- Discord/Telegram durable delivery adapter and worker tests

### 2. Tighten Discord delivery semantics around `visible_terminal_delivery`
Current behavior:
- `visible_terminal_delivery` can mean either a truly confirmed user-visible final delivery or an assumed/treated-as-visible terminal path depending on the branch
- this makes logs harder to trust during incident analysis

Prescription:
- split the concept into separate fields, for example:
  - `confirmed_surface_final_delivery`
  - `assumed_terminal_visibility`
- make logging use the precise field names so delivery forensics are unambiguous
- audit existing dashboards/tests that key off `visible_terminal_delivery`

Likely touch points:
- `src/codex_autorunner/integrations/discord/message_turns.py`
- log-based chat-surface lab scenarios and assertions

### 3. Remove duplicated Telegram durable-delivery abandonment helpers
Current behavior:
- the Telegram direct-delivery cleanup logic now exists in both:
  - `src/codex_autorunner/integrations/telegram/handlers/commands/execution.py`
  - `src/codex_autorunner/integrations/telegram/handlers/commands_runtime.py`

This is workable but creates drift risk.

Prescription:
[...]

PR #1554, opened 2026-04-22
[codex] Add direct-surface delivery ledger state

Git-on-my-level:
## What changed
- add a first-class durable ledger terminal state for direct surface delivery: `direct_surface_delivered`
- retain the original durable claim when the initial direct durable handoff fails on managed-thread turns, and pass that claim token through the shared managed-turn flow to Discord and Telegram
- replace the old post-send abandon path with shared direct-delivery bookkeeping that records either `direct_surface_delivered` after a successful visible fallback send or `retry_scheduled` if that fallback send fails
- add shared core and surface regressions covering the new terminal state, direct-send claim preservation, and worker/recovery exclusion of direct-delivered records

## Why
Issue #1553 called out a remaining race in the PMA duplicate-delivery fix: the surface could send a visible final reply and only abandon the durable record afterward, leaving a narrow window where the worker could still claim and replay the same finalized turn.

This change closes that race on the primary path by keeping the durable record claimed through the direct fallback send and then committing an explicit terminal ledger state when the direct surface delivery succeeds.

## Impact
- Discord and Telegram direct fallback sends now complete the same durable delivery obligation instead of abandoning it after the fact
- replay/recovery logic treats direct-surface-delivered records as terminal and non-deliverable
- managed-surface flows preserve the delivery claim token needed to finish direct fallback bookkeeping cleanly

## Validation
- aggregate commit lane passed locally, including strict mypy, full pytest, frontend build/tests, and chat-surface lab checks
- `.venv/bin/pytest -q tests/core/orchestration/test_managed_thread_delivery_ledger.py tests/core/orchestration/test_managed_thread_delivery_recovery.py`
- `.venv/bin/pytest -q tests/integrations/chat/test_managed_turn_runner.py`
- `.venv/bin/pytest -q tests/integrations/discord/test_discord_managed_thread_delivery_adapter.py`
- `.venv/bin/pytest -q tests/integrations/discord/test_message_turns_transient_progress.py -k durable_delivery`
- `.venv/bin/pytest -q tests/integrations/telegram/test_telegram_managed_thread_delivery_adapter.py`
- `.venv/bin/pytest -q tests/telegram_pma_managed_thread_support.py -k durable_delivery`

## Scope note
- this addresses the direct-delivery ledger-state slice of #1553
- the Telegram queue teardown warnings and the other follow-up items in #1553 remain separate work

### DIFF
--- a/src/codex_autorunner/core/orchestration/managed_thread_delivery.py
+++ b/src/codex_autorunner/core/orchestration/managed_thread_delivery.py
@@ -39,6 +39,7 @@ class ManagedThreadDeliveryState(str, Enum):
     DELIVERING = "delivering"
     RETRY_SCHEDULED = "retry_scheduled"
     DELIVERED = "delivered"
+    DIRECT_SURFACE_DELIVERED = "direct_surface_delivered"
     FAILED = "failed"
     ABANDONED = "abandoned"
     EXPIRED = "expired"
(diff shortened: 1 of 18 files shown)


Git-on-my-level:
@cursor address the codex review comments

Git-on-my-level:
@cursor fix failing ci

PR #1558, opened 2026-04-22
[codex] Show live progress for chat-bound review responses

Git-on-my-level:
## What changed

This PR makes queued GitHub review-response work visible in chat-bound threads when the agent actually wakes up and starts executing.

It adds a shared bound-chat live progress bridge for Discord and Telegram, supports editable progress outbox operations on both transports, and wires the managed-thread runtime so wake-up progress is shown and updated during execution.

It also fixes SCM notice routing so `Queued...` and `Started...` notices target the active bound chat surface directly, and it tightens the success lifecycle so the working anchor is only retired after final assistant delivery is enqueued.

## Why

The existing review-response path could be silent until the final turn arrived. That made it unclear whether another agent should be delegated, and created avoidable confusion when a second agent picked up work that had already been handled.

The root cause was twofold:

1. Wake-up execution happened later in the managed-thread queue worker, but there was no chat-visible working anchor at that layer.
2. Bound notice delivery relied on generic routing that could miss the active chat-bound thread.

## Impact

Users now get the same kind of live working indication for chat-bound review-response wake-ups that they already get for normal direct chat turns.

This applies to Discord and Telegram, with shared logic for the bound-chat progress lifecycle and transport-specific handling kept at the outbox layer.

## Validation

- Full repo validation lane via commit hook passed
- `7659 passed` in repo pytest suite
- Chat surface latency budget suite passed
- Focused regressions passed during development:
  - `tests/test_bound_chat_live_progress.py`
  - `tests/surfaces/web/routes/pma_routes/test_managed_thread_runtime.py`
  - `tests/test_telegram_outbox.py`
  - `tests/integrations/discord/test_outbox_retry.py`

## Notes

The branch includes both the original wake-up visibility fix and the follow-up refactor to make the bound-chat progress path immediate and to retire the progress anchor only after final delivery is handed off.


Git-on-my-level:
@codex review

Git-on-my-level:
@cursor review

Git-on-my-level:
Addressed the remaining PMA review threads on head `8aea78e6`.

- PMA finalization now preserves `backend_thread_id` by normalizing finalized/start/fallback values instead of dropping the caller-provided fallback.
- Queued PMA finalization returns the shared `ManagedThreadFinalizationResult` shape directly, so the queue worker and execution hooks consume one supported type.
- The stale startup-recovery-progress concern is covered by the existing durable final-delivery cleanup path; I added `test_startup_recovery_success_handoff_runs_delivery_cleanup_before_hooks` to pin that contract.
- The dead fallback plumbing around PMA finalization was removed as part of the normalization helper and shape cleanup.

Focused PMA/startup-recovery regressions passed locally, and the pushed head already cleared the aggregate lane in CI.

PR #1632, opened 2026-04-27
[codex] Unify ticket lint entrypoints

### LINKED ISSUE #1626: Unify CAR ticket lint entrypoints behind a single implementation
## Problem

We currently have two separate CAR ticket lint entrypoints:

- `.codex-autorunner/bin/lint_tickets.py`
- `.codex-autorunner/bin/ticket_tool.py lint`

They are not just two names for the same code path; they each carry their own lint implementation. That is already maintenance debt, and it creates avoidable drift risk for ticket filename validation, frontmatter rules, ticket_id handling, and error/reporting behavior.

During recent Discord ticket-browser investigation work, this duplication surfaced as unnecessary ambiguity about which lint command should be considered authoritative.

## Recommendation

Yes, we should unify them.

The repo should have exactly one real ticket-lint implementation. If we want to preserve both CLI entrypoints for compatibility, one of them should become a thin wrapper around the other shared implementation rather than owning its own copy of the logic.

## What I would like to see

1. Extract the lint behavior into one shared implementation.
   - Prefer a reusable Python function/module rather than copy-pasted script logic.
   - The implementation should own:
     - ticket filename validation
     - frontmatter parsing
     - agent/done/ticket_id validation
     - duplicate ticket_id validation
     - exit-code behavior

2. Keep both entrypoints only if they are true wrappers.
   - `.codex-autorunner/bin/lint_tickets.py` can stay as a convenience command if needed.
   - `ticket_tool.py lint` should call the same shared implementation.
   - No duplicated parsing/validation logic should remain in both files.

3. Make behavior intentionally identical.
   - Same success/failure exit codes
   - Same "no tickets" behavior
   - Same output shape and error reporting
   - Same handling of optional fix-up flags, if any remain supported

4. Add parity coverage.
   - Tests should prove that both entrypoints produce the same result against the same ticket directory.
   - Include at least:
     - valid queue
     - invalid frontmatter
     - duplicate ticket_id
     - empty ticket directory / missing tickets case

5. Document the preferred entrypoint.
   - Update AGENTS/docs/help text so contributors know which command is canonical.
   - If we intend to keep both long-term, say so explicitly.
   - If we intend to deprecate one later, note that path.

## Suggested acceptance criteria

- There is only one authoritative lint implementation in the codebase.
- Both CLI entrypoints, if retained, delegate to that shared implementation.
- Tests enforce behavioral parity between the entrypoints.
- Contributor-facing docs no longer imply two independent linters.

## Nice-to-have

- Consider moving ticket CLI/lint behavior into a small importable module under `src/codex_autorunner/tickets/` so the scripts in `.codex-autorunner/bin/` stay thin and easier to maintain.

Git-on-my-level:
## Summary
- move CAR ticket lint behavior into one shared portable implementation seeded as `.codex-autorunner/bin/_ticket_lint_impl.py`
- make both `.codex-autorunner/bin/lint_tickets.py` and `ticket_tool.py lint` thin wrappers over that shared implementation
- add entrypoint parity coverage and document `lint_tickets.py` as the canonical lint command while keeping `ticket_tool.py lint` as a compatibility alias

## Validation
- `python -m pytest -q tests/test_ticket_linter_cli.py tests/test_ticket_manager_cli.py tests/test_ticket_lint_entrypoint_parity.py tests/test_ticket_helper_script_sync.py tests/tickets/test_lint.py tests/test_agent_workspace_docs.py`
- full pre-commit validation gate via `git commit`, including repo-wide pytest, frontend build/tests, and chat-surface lab checks

Closes #1626


Git-on-my-level:
Addressed the outstanding review feedback and the aggregate/check failure in `03e31067`.

Changes in this update:
- fixed the portable fallback YAML parser so list items shaped like `- path: ...` are parsed as mappings, including `context:` entries with nested fields and single-line mapping items
- mirrored the parser fix into the generated helper source to keep the seeded scripts in sync
- exported `run_ticket_lint` as a public surface so the dead-code gate stops flagging the shared entrypoint that the generated wrappers import dynamically
- added regression coverage for the no-PyYAML fallback and entrypoint parity

Local validation passed before push:
- aggregate commit hook
- repo-wide strict mypy
- repo-wide pytest
- frontend build and JS tests
- chat-surface lab checks
- targeted `python scripts/deadcode.py --check`

Fresh CI is running on the new head now.

PR #1644, opened 2026-04-27
[codex] Fix PMA hygiene safe worktree purge

### LINKED ISSUE #1642: Add atomic car hub worktree purge command for safe full-lifecycle worktree removal
## Problem

There is no single CAR CLI operation that fully removes a worktree from all layers (hub registry, git metadata, disk). Today, cleanup requires 3 manual steps across 2 tools, each with failure modes that leave orphaned state:

1. `car hub worktree archive <id> --path <hub_root>` — removes from hub registry
2. `git -C <base_repo> worktree remove <path> --force` — removes from git worktree metadata
3. `rm -rf <path>` — removes directory from disk (raw shell command, unsafe)

This gap forces agents to hand-roll raw `rm -rf`, which causes cascading problems (CWD traps, confusing errors on failure).

## Failure modes observed during 2026-04-27 bulk cleanup

| Failure | Cause | Symptom |
|---------|-------|---------|
| Step 1 succeeds, ghost remains | manifest.yml not mutated | UI shows MISSING/UNINIT |
| Step 2 no-op, step 3 fails | Terminal CWD trapped in deleted dir | `FileNotFoundError` for wrong path |
| Steps 1+2 succeed, step 3 skipped | Assumption git remove cleans everything | Orphaned `.codex-autorunner/` dir remains |

### Chat-bound worktrees are NEVER auto-purged (intended behavior)

Chat-bound worktrees are protected unconditionally — even if corrupted (no `.git`, no source code, only stale orchestration DB). The safety gate is absolute: only `--force` with explicit human intent can touch them.

### Shell CWD trap (absorbed #1643)

When a shell's CWD is inside a directory that gets deleted, subsequent filesystem ops fail with confusing errors. Any disk removal must run in its own CWD-safe subprocess.

## Preferred Solution: Enhance `car pma hygiene --apply`

Rather than adding a new top-level `purge` command, extend the existing hygiene tool's apply mode to handle full-lifecycle worktree removal. This is more ergonomic — users already use `car pma hygiene` for classification; applying the cleanup should be the natural next step.

### Current behavior

```
$ car pma hygiene --apply --category automation --path <hub_root>
# → Cleans up subs/wakeups only (200 items processed)
# → Worktree archival is left as manual follow-up
```

### Proposed behavior

```
$ car pma hygiene --apply --category threads --path <hub_root>
[APPLY] Thread: codex-autorunner--dogfood-p3-8a2b246599
  [1/4] Archiving from hub registry... OK
  [2/4] Removing git worktree metadata... OK (pruned)
  [3/4] Removing directory from disk... OK (253MB reclaimed)
  [4/4] Verifying clean... OK
[APPLY] Complete: 9 worktrees purged, ~2.25GB reclaimed, 0 errors
```

### Safety properties

1. **CWD-safe subprocess** — spawns its own process with `cwd=/tmp`, immune to caller CWD state
2. **Respects chat_bound absolutely** — skips chat-bound worktrees with log note; only `--force` overrides
3. **Code-based deletion, not `rm -rf`** — use `shutil.rmtree()` or equivalent with:
   - Explicit error callback for permission failures (don't silently swallow)
   - Pre-flight size check and confirmation of what will be removed
   - Post-deletion verification that the directory is actually gone
4. **Per-step verification** — won't attempt disk removal if archive fails; reports exactly what happened at each stage
5. **`--dry-run` already exists** — reuse for thread category too; shows what would happen without side effects
6. **Idempotent** — re-running on already-purged worktree is a no-op with "already purged" message
7. **Respects existing safe/protected/needs-confirmation classification** — only applies cleanup to items classified as `safe`; needs-confirmation items are listed but skipped unless `--force`

### Why not a separate `purge` command?

- `hygiene --apply` is the established pattern — users classify first, then apply
- A new `purge` command duplicates the classification logic that `hygiene` already provides
- `--apply --category threads` keeps the mental model simple: one tool, two modes (inspect + act)
- Avoids CLI surface area growth for functionality that naturally extends an existing command

Git-on-my-level:
## Summary
- classify stale worktree-backed PMA thread hygiene candidates as safe or needs-confirmation at the worktree level
- route safe hygiene apply operations through the canonical worktree cleanup lifecycle instead of thread-only archiving
- harden leftover worktree directory removal with a CWD-safe subprocess and stop cleanup before manifest removal if disk verification fails

## Why
`car pma hygiene --apply --category threads` previously stopped at archiving the managed thread record. It did not purge the hub manifest entry, git worktree metadata, or leftover on-disk CAR state, which left ghost worktrees behind and pushed operators toward unsafe manual `rm -rf` cleanup.

## Validation
- full `core` validation lane from the commit hook
- `.venv/bin/python -m ruff check src/codex_autorunner/core/hub_worktree_manager.py src/codex_autorunner/core/pma_hygiene.py src/codex_autorunner/surfaces/cli/pma_cli.py tests/core/test_pma_hygiene.py tests/test_hub_supervisor.py`
- `.venv/bin/pytest -q tests/test_hub_single_owner_invariants.py -k 'archive_required_but_not_requested'`
- `.venv/bin/pytest -q tests/core/test_pma_hygiene.py tests/test_hub_supervisor.py -k 'pma_hygiene or leftover_car_state or stale_worktree'`

Closes #1642

### DIFF
--- a/src/codex_autorunner/core/hub_worktree_manager.py
+++ b/src/codex_autorunner/core/hub_worktree_manager.py
@@ -1,10 +1,12 @@
 from __future__ import annotations
 
+import json
 import logging
 import re
-import shutil
 import sqlite3
 import subprocess
+import sys
+import tempfile
 from pathlib import Path
 from typing import (
     TYPE_CHECKING,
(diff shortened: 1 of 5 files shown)


Git-on-my-level:
@cursor address the codex review comments and ensure make check passes

PR #1706, opened 2026-05-07
Trim cumulative managed-thread assistant output

Git-on-my-level:
## Summary
- trim Hermes cumulative terminal finalOutput at managed-thread finalization before delivery, transcript writes, timeline persistence, and execution result recording
- build durable prior-output prefixes from transcript history plus the previous completed execution so the fix works across old cumulative rows and future trimmed rows
- expose previous completed execution lookup through local and remote hub control-plane execution stores

## Investigation
- live pma-2 channel binding: Discord channel 1488827014600331415, managed thread 73e11d3c-1a77-42de-9e2f-e7a95624e8dc, backend thread 6c3071ed-b3cd-4a96-8623-c2a225b20f04
- orchestration DB showed assistant text growing from 908 to 6027 chars over turns 2-6
- cold traces showed stream deltas contained only the current answer, while terminal prompt/completed finalOutput contained prior assistant text plus the new answer

## Validation
- commit hook aggregate lane passed
- 8130 pytest tests passed
- strict mypy passed
- ruff passed
- frontend build/tests passed
- chat-surface latency budgets passed


Git-on-my-level:
Addressed the two latest Codex review threads in 11987cf17.

- Reworked transcript-prefix reconstruction so legacy cumulative transcript rows replace the accumulated prefix instead of being concatenated verbatim; newly trimmed single-turn rows are appended. This covers mixed histories like A, A+B, A+B+C, D.
- Removed the fixed 50-row cap for the trim-prefix transcript read. `TranscriptHistoryRequest(limit=0)` now means unbounded history, and managed-thread finalization uses that mode.
- Fixed the transcript lookup target kind to `thread_target`, matching how transcript mirrors are stored for managed threads.
- Added regression coverage for cumulative-row collapse, unbounded transcript-history requests, and unbounded hub transcript-history reads.

Validation: aggregate commit hook passed, including 8132 pytest tests, strict mypy, ruff, black check, frontend build/tests, and chat-surface latency budgets.

Git-on-my-level:
Rechecked the latest review state after the SCM notification. I do not see any new actionable review threads on the current head (`11987cf17`).

The two Codex review comments currently returned by GitHub are the prior threads on commit `e1fdc44`; both are now marked outdated after the previous update. No additional branch changes were needed.

PR #1710, opened 2026-05-10
[codex] Hide log output deltas from PMA chat

Git-on-my-level:
## Summary
- Hide persisted `output_delta` log lines from the visible PMA chat transcript.
- Keep commentary notices and intentional thinking/progress cards visible.
- Refresh the committed Web Hub static bundle.

## Root Cause
Command and file output is persisted as canonical timeline `intermediate` rows with `event_type: output_delta` and `intermediate_kind: log_line`. The frontend already suppressed assistant stream/final echo deltas, but did not suppress `log_line`, so raw tool stdout could render as a fully expanded chat message even though the tool call itself is represented by the collapsed tool trace UI.

## Validation
- `pnpm web:test -- pmaChat`
- `pnpm web:lint`
- `pnpm web:test`
- `pnpm run build`
- Commit hook web-ui lane: 8,522 pytest tests passed; 447 frontend tests passed; static asset check passed.


Git-on-my-level:
@cursor resolve merge conflicts for me

PR #1711, opened 2026-05-10
[codex] Improve PMA chat followups

Git-on-my-level:
## Summary
- make web chat sends return immediately via deferred managed-thread execution so follow-up messages are not blocked by a long-running first send
- add a compact queued-message strip above the composer sourced from the backend thread queue
- add interrupt affordances for the current draft and queued messages, using the existing managed-thread interrupt busy policy
- regenerate the committed Web Hub static bundle

## Validation
- pnpm web:test
- pnpm web:lint
- pnpm run build
- git commit hook web-ui lane: full pytest, mypy, web lint, static build, web tests, SPA shell/static checks

Note: the first commit attempt hit a transient Hypothesis too_slow health check in tests/contracts/memory/test_memory_store_contract.py; the targeted rerun passed, and the normal hook passed on retry.


Git-on-my-level:
Rechecked the latest review feedback. The three actionable threads are already addressed on the current head `18c799c9a`: failed-send draft restoration is version-gated, queued-turn interrupt preserves attachment intents, and pending attachments are cleared when send starts to avoid duplicate attachment sends. I marked those addressed threads resolved; current CI is green.

PR #1717, opened 2026-05-10
[codex] Fix worktree archive action visibility

Git-on-my-level:
## Summary

Fixes the repo index archive affordance for child worktree rows. The backend already exposes `has_car_state`, `unbound_managed_thread_count`, and chat-binding flags for worktrees, but the frontend child-row view model dropped those values and replaced them with defaults.

## Root Cause

`worktreeToNavChildRow()` hard-coded archive eligibility fields to false/zero. Detail pages and top-level worktree rows used the backend fields correctly, so the same worktree could show the Archive button on detail but not when nested under a repo.

## Changes

- Preserve worktree archive and cleanup-related raw flags on repo child worktree rows.
- Update component coverage so repo child rows render Archive when CAR state exists.
- Add view-model regression coverage for the preserved child-row flags.
- Refresh committed Web Hub static assets.

## Validation

- Focused Vitest: `pnpm --dir src/codex_autorunner/web_frontend exec vitest run src/lib/viewModels/repoWorktree.test.ts src/lib/components/RepoWorktreeViews.test.ts`
- Web lint/typecheck: `pnpm --dir src/codex_autorunner/web_frontend run lint`
- Commit hook web-ui lane: Python static checks, `mypy`, `pytest` (`8526 passed`), Web Hub lint, build, Vitest (`451 passed`), SPA shell check, and web static asset check.


Git-on-my-level:

PR #1727, opened 2026-05-11
[codex] Fix live progress folding in chat transcript

Git-on-my-level:
## Summary
- Fix live PMA progress rendering so tail-stream progress events that lack per-event managed_turn_id inherit the active run turn id.
- Merge same-turn Progress notice deltas into one trace card before the Work summary folds them, while keeping Commentary visible as a chat bubble.
- Refresh generated Web Hub static assets.

## Root cause
The frontend test fixture gave live progress events managed_turn_id, but actual tail-stream events put managed_turn_id on the run snapshot and not on each event. Those events could not be associated with the active turn, so they rendered as top-level progress bubbles instead of inside the Work toggle.

## Validation
- pnpm web:test -- pmaChat.test.ts
- pnpm web:lint (0 errors; existing warnings in PageHero, TicketViews, ScopedNewTicketPage)
- pnpm web:test (40 files, 467 tests)
- pnpm run build
- commit hook web-ui lane: 8,533 pytest tests passed; frontend lint/build/test passed; SPA shell/static asset checks passed


Git-on-my-level:

PR #1731, opened 2026-05-11
[codex] fix hermes pma tail profile streaming

Git-on-my-level:
## Summary
- Resolve managed-thread tail harnesses with the thread's `agent_profile` so profiled Hermes PMA chats attach to the same live supervisor used for execution.
- Add a regression test proving profiled Hermes tail snapshots call `harness_factory("hermes", "m4-pma")`.

## Root Cause
Web PMA deferred execution preserved the Hermes profile when starting the turn, but `/tail/events` rebuilt the live harness with only `agent_id="hermes"`. For `m4-pma` chats this selected a different Hermes supervisor, so the SSE stream crashed with `Unknown Hermes turn 'turn-1'` and the browser fell back to polling.

## Validation
- Reproduced locally against real Hermes `m4-pma`: `/tail/events` closed after one state frame with `HermesSupervisorError: Unknown Hermes turn 'turn-1'`.
- Re-ran locally against real Hermes after the fix: `/tail/events` stayed connected through backend binding handoff and emitted live `tail`/`timeline` frames at 9.5s.
- `.venv/bin/python -m pytest -q tests/surfaces/web/routes/pma_routes/test_tail_stream_characterization.py tests/test_managed_threads_tail.py`
- `.venv/bin/python -m pytest -q tests/agents/hermes/test_hermes_supervisor.py tests/agents/hermes/test_hermes_supervisor_official_prompt_hang.py`
- Commit hook web-ui lane: black, ruff, import/contract checks, mypy, 8536 pytest tests, web lint/build/test, SPA shell check.

### DIFF
--- a/src/codex_autorunner/surfaces/web/routes/pma_routes/tail_stream.py
+++ b/src/codex_autorunner/surfaces/web/routes/pma_routes/tail_stream.py
@@ -124,14 +124,41 @@ def resolve_resume_after(
     return parsed
 
 
-def _managed_thread_harness(service: Any, agent_id: str) -> Any:
+def _managed_thread_harness(
+    service: Any, agent_id: str, profile: Optional[str] = None
+) -> Any:
     factory = getattr(service, "harness_factory", None)
     if not callable(factory):
         return None
     try:
+        if profile:
+            try:
+                return factory(agent_id, profile)
+            except TypeError as exc:
+                if "positional argument" not in str(exc):
+                    raise
+                return factory(agent_id)
         return factory(agent_id)
     except Exception:  # intentional: dynamic harness factory - exception types unknown
         return None
+
+
+def _managed_thread_agent_profile(thread: Any) -> Optional[str]:
+    profile = normalize_optional_text(getattr(thread, "agent_profile", None))
+    if profile:
+        return profile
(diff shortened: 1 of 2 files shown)


Git-on-my-level:
Addressed the review feedback around the legacy factory fallback.

What changed:
- Moved the profiled `factory(agent_id, profile)` retry logic into `_managed_thread_harness()` at the actual factory invocation level, so `TypeError` from old one-argument factories is no longer swallowed before fallback can run.
- Kept the unprofiled path calling `_managed_thread_harness(service, agent_id)` so existing monkeypatched/legacy two-argument helper usage stays compatible.
- Added `test_tail_snapshot_falls_back_for_legacy_harness_factory` to prove profiled threads still get a valid harness from legacy `harness_factory(agent_id)` services.

Validation:
- `.venv/bin/python -m pytest -q tests/surfaces/web/routes/pma_routes/test_tail_stream_characterization.py tests/test_managed_threads_tail.py` -> 42 passed
- Amend hook web-ui lane passed again: black, ruff, import/contract checks, mypy, 8536 pytest tests, web lint/build/test, SPA shell check.

Pushed updated commit: `4c888ddf8`.

PR #1778, opened 2026-05-13
[codex] Model OpenCode turn lifecycle explicitly

### LINKED ISSUE #1774: Model OpenCode turns with an explicit lifecycle state machine
## Context

A first-turn PMA chat regression exposed a lifecycle ambiguity in the OpenCode harness: `prompt_async` can return before any SSE/progress events arrive. The harness had a fast path where command completion plus empty immediate message recovery could finalize the managed turn as successful with empty assistant output. A later turn then showed that OpenCode had continued producing real assistant events after CAR already committed the first turn as empty.

The immediate fix keeps collecting when the early prompt-response recovery is empty. That prevents this specific race, but the architecture still has two competing concepts of completion embedded in `wait_for_turn()`.

## Architecture smell

OpenCode turn handling currently mixes these signals inline:

- HTTP command completion: `prompt_async` returned.
- Runtime turn completion: assistant output and terminal/idle lifecycle were observed.
- Snapshot recovery: `list_messages()` produced recoverable assistant text.
- Progress synthesis: polling messages to create progress events while SSE is sparse.

The bug came from allowing transport acknowledgment to short-circuit runtime observation. `prompt_async` returning should mean only "the backend accepted or queued the prompt". It should not mean the assistant turn is done.

## Current gaps

- `wait_for_turn()` races `command_task` and `collect_task`, then directly decides how to combine them. This makes it easy for the wrong completion signal to win.
- Empty successful turns are too easy to represent: `status=ok`, `assistant_text=""`, no error. That should be a degraded terminal state, not normal success.
- Message snapshot recovery is mixed into lifecycle control. Recovery should enrich output after a lifecycle decision, not decide lifecycle by itself unless it has trusted final output.
- There is no typed distinction between accepted, observing, terminal-observed, snapshot-recovered, failed, and empty-terminal.
- The current policy relies on negative evidence like "no progress events published and no preconnected event seen", which is brittle.

## Ideal model

Introduce an explicit OpenCode turn lifecycle state machine. Suggested states:

```python
class OpenCodeTurnLifecycleState(Enum):
    ACCEPTED = "accepted"
    OBSERVING = "observing"
    TERMINAL_OBSERVED = "terminal_observed"
    SNAPSHOT_RECOVERED = "snapshot_recovered"
    EMPTY_TERMINAL = "empty_terminal"
    FAILED = "failed"
````

Suggested typed result:

```python
@dataclass(frozen=True)
class OpenCodeTurnLifecycleResult:
    state: OpenCodeTurnLifecycleState
    assistant_text: str
    terminal_signal: str | None
    output_source: Literal[
        "event_stream",
        "prompt_response",
        "messages_snapshot",
        "none",
    ]
    command_completed: bool
    terminal_observed: bool
    snapshot_recovered: bool
    error: str | None
    raw_events: list[dict[str, object]]
    evidence: dict[str, object]
````

## Target invariants

- `prompt_async` returning transitions at most to `ACCEPTED`; it never finalizes a turn by itself.
- `wait_for_turn()` returns `ok` only when assistant text came from a trusted terminal observation or snapshot recovery.
- Empty assistant output is explicit: `EMPTY_TERMINAL` if a real terminal signal was observed with no output; otherwise keep observing until timeout/failure.
- `list_messages()` snapshot recovery can produce `SNAPSHOT_RECOVERED` only when it finds assistant text for the active turn/session.
- Sparse SSE and preconnected-stream races are handled by lifecycle policy, not by ad hoc early-cancel branches.
- Progress synthesis can publish UI events without owning final-turn completion semantics.

## Proposed implementation steps

1. Extract the command-vs-collector coordination from `OpenCodeHarness.wait_for_turn()` into a small lifecycle coordinator/state machine.
2. Encode transitions for command completion, stream event, stream idle/terminal, snapshot text found, stream timeout, and transport failure.
3. Make empty prompt-response recovery continue observing instead of producing success.
4. Make snapshot recovery typed and require assistant text before it can finalize as recovered.
5. Return a typed lifecycle result from the coordinator, then adapt it into `TerminalTurnResult` at the harness boundary.
6. Add regression tests for first-turn races: command returns before SSE, preconnected stream exhausts early, `list_messages()` initially empty, then real assistant event arrives.
7. Add tests for true empty-terminal behavior so intentional no-output turns remain observable and diagnosable.
8. Add telemetry fields for lifecycle state, terminal signal, output source, command completion timing, and snapshot recovery timing.

## Acceptance criteria

- No production path treats `prompt_async` return as terminal turn completion.
- `status=ok` with empty `assistant_text` is impossible unless the lifecycle state explicitly records `EMPTY_TERMINAL`.
- First-turn OpenCode chats cannot commit an empty successful assistant turn while the backend is still producing output.
- Tests cover delayed SSE after command completion, empty early snapshot recovery, terminal idle with final text, terminal idle without final text, and late stream disconnect after terminal output.
- Logs/telemetry make the final lifecycle state and output source obvious for future debugging.

## Related files

- `src/codex_autorunner/agents/opencode/harness.py`
- `src/codex_autorunner/agents/opencode/runtime.py`
- `src/codex_autorunner/agents/opencode/stream_lifecycle.py`
- `src/codex_autorunner/agents/opencode/output_assembly.py`
- `tests/agents/opencode/test_opencode_harness.py`
- `tests/test_opencode_runtime.py`

Git-on-my-level:
## Summary
- extract OpenCode turn lifecycle policy into `turn_lifecycle.py` instead of embedding it in the harness
- make `prompt_async` completion a transport acceptance signal only; it can no longer supply final assistant output
- remove brittle prompt-response fallback paths from runtime, backend, file chat, and PMA chat execution
- make output source explicit as event stream, message snapshot, or none; snapshot recovery is now typed instead of inferred from polling side effects
- log lifecycle state, terminal signal, output source, command timing, and snapshot recovery evidence
- cover delayed SSE after prompt acceptance, snapshot recovery, empty terminal, and command-failure ordering cases

Fixes #1774

## Validation
- commit hook aggregate lane: black, ruff, import/contract checks, repo-wide mypy, 8702 pytest tests, Web UI lab, frontend lint/build, Vitest
- targeted: `.venv/bin/python -m pytest tests/agents/opencode/test_turn_lifecycle.py tests/agents/opencode/test_opencode_harness.py tests/agents/opencode/test_output_assembly.py tests/test_opencode_runtime.py tests/test_opencode_run_prompt.py tests/test_opencode_backend_lifecycle.py tests/test_opencode_backend_streaming.py tests/test_opencode_backend_approvals.py -q`
- targeted: `.venv/bin/python -m ruff check ...` on edited OpenCode/backend/web route files
- targeted: `.venv/bin/python -m mypy src/codex_autorunner/agents/opencode src/codex_autorunner/adapters/agents/opencode_backend.py src/codex_autorunner/surfaces/web/routes/file_chat_routes/execution_agents.py src/codex_autorunner/surfaces/web/routes/pma_routes/chat_runtime_execution.py`


Git-on-my-level:
Addressed the review feedback in `666f38bcf`.

- Preserved command/prompt response errors in the lifecycle coordinator so a command acknowledgement that reports an error fails with that specific error instead of waiting for a generic stream timeout.
- Added regression coverage in `tests/agents/opencode/test_turn_lifecycle.py` for that path.
- Kept prompt-response text out of final assistant output by design. The lifecycle still only accepts final text from runtime observation or trusted message snapshot recovery, which avoids reintroducing the brittle compatibility path this PR is removing.

Validation: commit hook aggregate lane passed: black, ruff, import/contract checks, repo-wide mypy, 8703 pytest tests, Web UI lab, frontend lint/build, and Vitest.

Git-on-my-level:
Rechecked the latest PR review state after the new-feedback ping.

I do not see any newer review comments or unresolved current-diff threads after the prior update in `666f38bcf`. The only review threads currently returned by GitHub are the two older comments on `9a19aab4`; both are now outdated against the current diff and were already addressed by preserving command-response errors while keeping prompt-response text out of the final output.

No branch update was needed because the working tree is clean and already matches `origin/fix/opencode-turn-lifecycle-1774`.

PR #1824, opened 2026-05-16
[codex] optimize Web UI page load windows

Git-on-my-level:
## Summary

- add a Playwright-based Web UI page-load profiler for fixture, live hub, and explicit URL modes
- reduce initial chat and repo/worktree read-model windows from 200 rows to 50 rows
- remove an unused root layout data import so shared layout startup avoids chat-index session code

## AutoOptimize Results

- ran blessed.autooptimize for 5 iterations against the main Web UI routes
- baseline p75 route-ready time: 985.972 ms
- best recorded AutoOptimize iteration: 954.193 ms
- final kept-state profile after test-aligned store defaults: 903.124 ms p75
- real hub-root final profile: 1010.168 ms p75; /repos still had a 4904.72 ms outlier worth follow-up

## Validation

- commit aggregate guardrails passed
- black, ruff, import-boundary checks
- hub interface/destination/keyword contract checks
- mypy strict repo-wide
- pytest: 8885 passed
- Web UI lab fast check: 26 scenarios passed
- pnpm web:lint
- pnpm web:build
- pnpm web:test: 643 passed, 2 skipped
- Web Hub SPA shell check
- car apps run blessed.autooptimize validate-state

### DIFF
--- a/scripts/profile_web_page_load_times.py
+++ b/scripts/profile_web_page_load_times.py
@@ -1,0 +1,402 @@
+#!/usr/bin/env python3
+"""Measure Web UI route ready times against a seeded or live hub."""
+
+from __future__ import annotations
+
+import argparse
+import json
+import os
+import shutil
+import statistics
+import sys
+import tempfile
+import time
+from pathlib import Path
+from typing import Any
+from urllib.error import URLError
+from urllib.request import urlopen
+
+REPO_ROOT = Path(__file__).resolve().parents[1]
+SRC_ROOT = REPO_ROOT / "src"
+if str(REPO_ROOT) not in sys.path:
+    sys.path.insert(0, str(REPO_ROOT))
+if str(SRC_ROOT) not in sys.path:
+    sys.path.insert(0, str(SRC_ROOT))
+
+import web_ui_screens
+from codex_autorunner.bootstrap import seed_hub_files
+from codex_autorunner.core.utils import atomic_write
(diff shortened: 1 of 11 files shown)


Git-on-my-level:
Addressed the latest review feedback and pushed `0c7f16474`.

- Kept the Bugbot same-request fix: `sameChatIndexRequest` now uses the 50-row default, matching window normalization.
- Restored the repo/worktree index loader default to 200 so `/repos` and `/worktrees` do not truncate discovery to 50 rows. The smaller 50-row window remains available only when a caller passes `limit: 50` explicitly.
- Changed the chat unread chip to use the backend aggregate unread counter as a floor, so unread chats outside the first 50-row client window do not get under-reported.
- Added focused coverage for the repo/worktree default versus explicit 50-row limit, and for unread chip counts when the first chat window is smaller than the full result set.

Validation after the patch:

- `pnpm --filter @codex-autorunner/web-hub test -- src/lib/data/readModelLoaders.test.ts src/routes/chats/page.test.ts src/lib/data/chatIndexSession.test.ts`
- `pnpm web:lint`
- `pnpm web:test`
- `pnpm --filter @codex-autorunner/web-hub build`
- Commit hook passed the normal `web-ui` lane on retry, including guardrails, mypy, full pytest (`8885 passed`), Web UI lab, lint/build/test, and SPA shell check.

PR #1830, opened 2026-05-16
[codex] Move Web Chat toward turn-structured detail projection

Git-on-my-level:
## Summary

- enforce turn-structured transcript ordering so live progress renders under the user message for the active turn
- stabilize the new-chat draft-to-committed handoff so the detail pane does not disappear while the chat index catches up
- add the first chat detail read-model projection contract and migration docs for the longer-term single-projection architecture

## Root Cause

- backend transcript assembly merged durable timeline rows with live tail rows and sorted by generic row keys, so progress rows could surface before the turn user message
- frontend active detail depended on the active chat being present in the chat index; clearing the local draft before the committed thread appeared made `activeChat` temporarily null

## Validation

- `python3 .codex-autorunner/bin/lint_tickets.py`
- `.venv/bin/python -m pytest tests/surfaces/web/routes/pma_routes/test_managed_thread_timeline_route.py tests/surfaces/web/test_hub_chat_read_model_routes.py tests/contracts/web/test_read_model_contracts.py`
- `pnpm web:test`

### DIFF
--- a/docs/architecture/web-ui-read-model-contracts.md
+++ b/docs/architecture/web-ui-read-model-contracts.md
@@ -226,6 +226,18 @@ Chat detail:
 - Patch: `ChatDetailPatchEvent`.
 - Carries thread metadata, visible timeline window, queue summary, artifacts,
   and optimistic reconciliation ids (`clientMessageId`, `backendMessageId`).
+- Timeline items carry backend-authored turn and ordering fields:
+  `managedTurnId`, `orderKey`, `section`, and `sectionOrder`. The frontend may
+  sort visible items by `orderKey` for idempotent patch replay, but it must not
+  infer turn structure from timestamps, `/tail`, or `/timeline`.
+- Compatibility bridge: the snapshot route is authoritative at
+  `/hub/read-models/chats/{chatId}` and the typed detail stream lives at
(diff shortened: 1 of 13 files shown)


Git-on-my-level:
Addressed the latest review feedback and pushed `22397132f`.

- Kept the Bugbot turn-anchor frontend sort and extended it to respect the chat-detail `sectionOrder`/section fallbacks, with regressions for both snapshot and patch application where same-turn progress has an earlier order key than the user row.
- Fixed the chat-detail patch stream cursor handling so batches filtered down to zero events still advance the cursor, preventing a quiet target chat from re-reading unrelated batches forever.
- Added a backend regression that walks past an unrelated chat batch and reaches the target chat event.

Validation:

- `pnpm --filter @codex-autorunner/web-hub test -- src/lib/data/readModelStore.test.ts`
- `.venv/bin/python -m pytest tests/surfaces/web/test_hub_chat_read_model_routes.py::test_hub_read_models_chat_detail_patch_cursor_skips_unrelated_batches tests/surfaces/web/test_hub_chat_read_model_routes.py::test_hub_read_models_chat_detail_patch_stream_is_repairable`
- Commit hook with full web-ui lane and repo pytest: `8903 passed, 10 warnings`; `web:test` also passed (`651 passed, 2 skipped`).

PR #1850, opened 2026-05-17
[codex] Fix ticket loop guard dirty diff fingerprint

### LINKED ISSUE #1849: Ticket-flow loop guard falsely pauses dirty multi-turn tickets
## Summary
Ticket flow can falsely pause with `Ticket loop detected (no repo diff change)` on long-running tickets that make real edits across multiple turns while the worktree is already dirty.

The loop guard currently fingerprints only `HEAD + git status --porcelain`. Once a ticket has modified a set of files, later edits to those same files do not change porcelain status, so real content changes can be counted as consecutive no-op turns.

## Observed Example
Repo/worktree: `codex-autorunner--discord-1`
Run: `152b854d-8e6d-49da-b824-e2929e0171b6`
Ticket: `.codex-autorunner/tickets/TICKET-140-pma-automation-migration.md`

The run paused at dispatch `0012` with:

```
Ticket loop detected (no repo diff change)
Consecutive no-change turns: 2
```

But dispatches `0008` through `0011` reported real implementation work and validation, and the worktree had a large uncommitted diff touching PMA automation and unified automation files.

## Root Cause
The no-change detector compares `repo_fingerprint_before == repo_fingerprint_after`, where the fingerprint is effectively:

```
HEAD
$(git status --porcelain)
```

Relevant code paths from the current checkout:

- `src/codex_autorunner/tickets/runner_execution.py`
  - `capture_git_state_after()` captures `git status --porcelain`
  - `compute_loop_guard()` treats identical fingerprints as no repo change
- `src/codex_autorunner/tickets/runner_post_turn.py`
  - `reconcile_post_turn()` pauses when `should_pause_for_loop()` reaches the threshold
  - `get_repo_fingerprint()` also uses `HEAD + porcelain status` for pause context

This is insufficient for dirty multi-turn work. If `src/foo.py` is already modified, another edit to `src/foo.py` leaves porcelain as `M src/foo.py`, so the fingerprint is unchanged even though the diff content changed.

## Expected Behavior
The loop guard should only count a turn as no-change when actual repository content did not change. Real edits to already-modified files should reset the no-change count.

## Suggested Fix
Use a content-sensitive fingerprint for dirty worktrees, for example one of:

- hash of `git diff --full-index --binary` plus staged diff and untracked file content metadata
- hash of `git diff --stat`/`--numstat` plus full diff for modified paths
- a helper that captures HEAD, index diff, worktree diff, and untracked file hashes

Also reset loop guard state on successful checkpoint/commit and ticket transitions.

## Why This Matters
Large tickets often require several turns before they are ready to mark done and commit. With the current fingerprint, those tickets can pause repeatedly even while the agent is making progress, which makes ticket flow appear to stop after every substantial ticket or after repeated turns on the same dirty ticket.

Git-on-my-level:
## Summary

Fixes #1849 by making ticket-flow loop guard fingerprints content-sensitive instead of relying only on `HEAD + git status --porcelain`.

The runner now includes staged diff, worktree diff, and untracked file content fingerprints in the repository snapshot used before and after each turn. Pause context also uses the same content-sensitive fingerprint helper.

## Impact

Long-running dirty ticket flows no longer accumulate false no-change turns when an agent keeps editing files that were already modified before the turn.

## Validation

- `.venv/bin/python -m pytest tests/tickets/test_runner_state_transitions.py::TestGetRepoFingerprint tests/tickets/test_runner_invariants.py::TestLoopGuardInvariants`
- `.venv/bin/python -m pytest tests/tickets/test_runner_state_transitions.py tests/tickets/test_runner_invariants.py tests/tickets/test_ticket_runner.py -q`
- commit hook aggregate validation: `8948 passed`, mypy clean, Ruff clean, Web UI lint/build/tests clean

### DIFF
--- a/src/codex_autorunner/core/flows/controller.py
+++ b/src/codex_autorunner/core/flows/controller.py
@@ -6,7 +6,7 @@ from pathlib import Path
 from typing import Any, AsyncGenerator, Callable, Dict, Optional, Set
 
 from ...manifest import ManifestError, load_manifest
-from ..git_utils import GitError, run_git
+from ..git_utils import git_content_sensitive_repo_fingerprint
 from ..lifecycle_events import LifecycleEventEmitter, LifecycleEventType
 from ..state_roots import resolve_hub_manifest_path
 from ..utils import find_repo_root
@@ -310,17 +310,7 @@ class FlowController:
         repo_root = self._repo_root()
         if repo_root is None:
             return None
-        try:
-            head_proc = run_git(["rev-parse", "HEAD"], cwd=repo_root, check=True)
-            status_proc = run_git(["status", "--porcelain"], cwd=repo_root, check=True)
-            head = (head_proc.stdout or "").strip()
-            status = (status_proc.stdout or "").strip()
-            if not head:
-                return None
-            return f"{head}\n{status}"
-        except GitError:
-            _logger.exception("Failed to get git state")
-            return None
+        return git_content_sensitive_repo_fingerprint(repo_root)
 
     def _has_new_user_reply_signal(
         self, *, run_id: str, input_data: dict[str, Any]
--- a/src/codex_autorunner/core/git_utils.py
+++ b/src/codex_autorunner/core/git_utils.py
@@ -2,6 +2,8 @@
 Centralized Git utilities for consistent git operations across the codebase.
 """
 
+import hashlib
+import os
 import subprocess
 from contextlib import contextmanager
 from pathlib import Path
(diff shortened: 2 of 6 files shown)


Git-on-my-level:
Addressed the latest review feedback in a1c6ea358.\n\n- Moved the content-sensitive repo fingerprint helper into core git utilities and updated FlowController._repo_fingerprint() to use the same format as ticket pause contexts, so resume gating compares compatible fingerprints.\n- Switched the untracked-file scan to binary git output and decode paths with filesystem decoding, avoiding UTF-8 decode crashes for unusual untracked filenames.\n- Kept CAR's untracked .codex-autorunner runtime files out of the content hash so internal flow DB writes do not count as user repo changes.\n\nValidation:\n- Focused loop/resume tests: 14 passed\n- Affected runner/resume files: 145 passed\n- Commit hook aggregate validation: Black/Ruff/import checks/mypy clean, pytest 8949 passed, Web UI lint/build/tests clean

PR #1880, opened 2026-05-20
[codex] Speed up automations workspace load

Git-on-my-level:
## Summary

- Add a backend-owned `/hub/read-models/automations/workspace` snapshot for the automations screen.
- Batch automation schedules, latest jobs, and job counts instead of doing per-rule enrichment queries in the primary overview path.
- Update `/automations` to render from the workspace snapshot first, then hydrate agent controls independently.
- Remove the legacy `/hub/repos` first-paint dependency from the automations route.

## Root Cause

The automations page was slow because it waited for three independent calls before leaving global loading: `/hub/automations`, `/hub/pma/agents`, and legacy `/hub/repos`. The automations payload itself was fast, but `/hub/repos` can be seconds slower because it performs broad repo topology/enrichment work. The backend overview also had a future scaling issue: each automation row separately loaded schedules, latest jobs, and job counts.

## Architecture

The automations screen now has a screen-shaped workspace read model with automations, summary counters, presets, lightweight target options, agent defaults, and a generation timestamp. The frontend uses that snapshot for first paint and treats agent/model catalog loading as local control hydration rather than a page-level blocker.

## Timing Evidence

Live hub before reload/current old endpoints:

- `/hub/automations`: HTTP 200, `total=0.093257`, size 99,703 bytes.
- `/hub/repos`: HTTP 200, `total=21.610828`, size 241,464 bytes.

Current-code in-process TestClient against the hub root:

- `/hub/read-models/automations/workspace`: HTTP 200, `total=0.029520`, size 106,418 bytes.
- `/hub/automations`: HTTP 200, `total=0.022683`, size 99,703 bytes.

The running hub had not reloaded this branch yet, so the new route returned 404 from the already-running process; the TestClient measurement exercises the current code directly.

## Validation

Targeted checks run manually:

- `.venv/bin/python -m pytest tests/surfaces/web/test_hub_automation_routes.py` passed: 10 passed.
- `pnpm web:test -- automations` passed: 1 file, 4 tests.
- `pnpm web:lint` passed: 0 Svelte errors/warnings.

Pre-commit selected lane also passed:

- black check.
- ruff.
- import boundaries.
- hub interface contracts.
- destination contract drift.
- keyword contracts.
- migration observability sync.
- mypy over `src/codex_autorunner`.
- broad pytest: 9,204 passed.
- Web UI lab fast scenarios: 27 passed.
- Web lint.
- Web build.
- full Web frontend tests: 780 passed, 2 skipped.
- Web Hub SPA shell check.

## Review Notes

I reviewed the staged implementation before publishing and did not find a blocking issue. Remaining operational note: the live hub needs to reload before the new endpoint can be timed via the running server process.

### DIFF
--- a/src/codex_autorunner/core/automation/product.py
+++ b/src/codex_autorunner/core/automation/product.py
@@ -213,7 +213,7 @@ def automation_store(hub_root: Path) -> AutomationStore:
 def automation_overview(store: AutomationStore, *, limit: int = 100) -> dict[str, Any]:
     take = max(1, min(int(limit or 100), 500))
     rules = store.list_rules()[:take]
-    rows = [automation_row(store, rule) for rule in rules]
+    rows = automation_rows(store, rules, recent_job_limit=25)
     summary = {
         "total": len(rows),
         "active": sum(1 for row in rows if row["enabled"]),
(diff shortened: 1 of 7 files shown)


Git-on-my-level:
Follow-up review pass: fetched thread-aware review state and flat PR/review comments. The only review thread present is the prior preset target picker comment; it is now outdated and was addressed by `5c9c527e8`, which restricts preset creation to base repo options. I found no additional active review comments to patch in this pass, so the branch remains at `5c9c527e8`.

PR #1926, opened 2026-05-22
fix(chat): reject foreign model on managed-thread turns before dispatch

### LINKED ISSUE #1854: PMA managed-thread silently accepts a foreign/invalid model and returns empty output
## Summary

A PMA managed-thread turn handed a model that does not belong to the chat's agent runs to completion, records `status = ok`, and stores an **empty `assistant_text`**. The agent silently produces no response and nothing surfaces as an error.

## Evidence

Found while investigating Hermes chat `5dbc28` not responding. The frontend was sending `model=gpt-5.4-mini` (a **Codex** model) on turns of a **Hermes** chat (`orch_thread_targets.agent_id = hermes`). `orch_thread_executions` for that thread:

| turn | model_id | assistant_text length |
|---|---|---|
| 1-4 | *(none)* | 58, 25, 754, 836 |
| 5-7 | `gpt-5.4-mini` | **0** |

The turns with the foreign model all completed as `status = ok` with zero assistant output. No error was raised, logged, or shown to the user.

## Why this matters

The frontend bug that sent the wrong model has been fixed (selector desync on URL-deep-link — chat opened before its summary loaded stayed on the hub default agent). But the backend has **no guard**: any caller (web, Discord, Telegram, CLI, a future regression) that passes a model the agent cannot use gets the same silent dead turn. A turn that produced nothing should never report `ok`.

## Expected behavior

When a turn specifies a model that is not in the resolved agent's model catalog (or that the agent cannot use), the orchestration layer should either:

- reject the turn with a clear validation error before execution, or
- fall back to the agent's default model and record that it did so,

and in any case **never record `status = ok` with empty `assistant_text`** for a turn that produced no output.

## Pointers

- `src/codex_autorunner/adapters/chat/managed_thread_turns.py` — turn finalization (`status`/`assistant_text` persistence)
- `src/codex_autorunner/core/orchestration/` — managed-thread execution path
- `orch_thread_executions.model_id` — where the foreign model lands

## Repro

1. Open a Hermes PMA chat by URL before the chat index loads (hard refresh of `/chats/<id>`).
2. Send a message — pre-fix, the composer sends a Codex model id.
3. Turn completes `ok`, transcript shows no assistant reply.

tmchow:
## Summary

A managed-thread turn that specified a model the resolved agent cannot use was accepted, ran to completion as `status=ok`, and stored an empty `assistant_text`. This rejects such a turn before it is dispatched, so the foreign model never reaches the agent session.

## Why this matters

> a single foreign-model turn poisons the whole chat. The guard should reject the foreign model **before** the turn is dispatched to the agent, so it never reaches the agent session at all.
>
> — @Git-on-my-level on #1854

Hermes chat `5dbc28` went dead because the frontend sent a Codex model (`gpt-5.4-mini`) on a Hermes chat. Turns 5-7 completed `ok` with zero assistant output and no error surfaced anywhere. Worse, Hermes (ACP) pins the bad model onto its session, so every later turn reuses the dead model and the chat stays bricked until a new chat is started or the session JSON is hand-edited.

The frontend selector bug is already fixed, but the backend had no guard. Any caller (web, Discord, Telegram, CLI, a future regression) that passes an unusable model gets the same silent dead turn.

The reported case is structural: Hermes has no `model_listing` capability, so it has no caller-selectable model catalog, and any non-empty model on such a turn is foreign by definition. The guard keys on that capability rather than fetching a live model catalog, so it costs nothing on the dispatch path and works for the reported agent (whose harness cannot list models at all).

## Changes

- Reject in `prepare_thread_execution` after the agent definition is resolved and before `create_execution`, so the foreign model never lands in `orch_thread_executions` and never reaches the ACP session.
- The check fires only when a non-empty model is set on an agent missing `model_listing`. Agents that support model selection (codex, opencode) are unaffected.
- `ForeignModelError` is exported from `core.orchestration` so callers can catch it.

## Testing

- `test_prepare_rejects_foreign_model_for_non_model_listing_agent`: a model on a Hermes thread raises `ForeignModelError`, the harness never starts a turn, and no execution row is created.
- `test_prepare_allows_model_for_model_listing_agent`: a `model_listing` agent still accepts a caller-specified model.
- Full orchestration suite passes (1091 tests).

This adds the pre-dispatch guard. It does not touch the frontend selector that originally sent the wrong model (already fixed per #1854).

Fixes #1854

[Compound Engineering](https://github.com/EveryInc/compound-engineering-plugin)

### DIFF
--- a/src/codex_autorunner/core/orchestration/__init__.py
+++ b/src/codex_autorunner/core/orchestration/__init__.py
@@ -222,7 +222,7 @@ if TYPE_CHECKING:
         build_surface_orchestration_ingress,
         get_surface_orchestration_ingress,
     )
-    from .thread_service import HarnessBackedOrchestrationService
+    from .thread_service import ForeignModelError, HarnessBackedOrchestrationService
     from .thread_store_adapter import (
         ManagedThreadExecutionStore,
     )
@@ -232,6 +232,10 @@ _LAZY_EXPORTS = {
     "FlowBackedOrchestrationService": (
         ".flow_service",
         "FlowBackedOrchestrationService",
+    ),
+    "ForeignModelError": (
+        ".thread_service",
+        "ForeignModelError",
     ),
     "HarnessBackedOrchestrationService": (
         ".thread_service",
@@ -323,6 +327,7 @@ __all__ = [
     "ExecutionRecord",
     "FlowBackedOrchestrationService",
     "FlowTarget",
+    "ForeignModelError",
     "FreshConversationRequiredError",
     "HarnessBackedOrchestrationService",
     "ManagedThreadDeliveryAttemptResult",
--- a/src/codex_autorunner/core/orchestration/thread_service.py
+++ b/src/codex_autorunner/core/orchestration/thread_service.py
@@ -55,6 +55,24 @@ logger = logging.getLogger("codex_autorunner.core.orchestration.service")
 HarnessFactory = Callable[..., RuntimeThreadHarness]
 
 
+class ForeignModelError(ValueError):
+    """Raised when a turn specifies a model the resolved agent cannot use.
+
+    An agent without the ``model_listing`` capability (e.g. Hermes/ACP) has no
+    caller-selectable model catalog, so any non-empty model on such a turn is
+    foreign. The turn is rejected before dispatch so the model never reaches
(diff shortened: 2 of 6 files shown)


Git-on-my-level:
@cursor review

Git-on-my-level:
@codex review

PR #1962, opened 2026-05-27
[codex] Fix OpenCode question tool turn handling

Git-on-my-level:
## Summary
- Recognize OpenCode `tool: question` message parts as user-input requests when they are pending.
- Treat already-failed OpenCode question tool parts as fatal turn errors so nonterminal progress text cannot be shown as a completed final answer.
- Add protocol and collector regression coverage for the deployed failure shape.

## Validation
- `.venv/bin/python -m pytest -q tests/agents/opencode/test_turn_runtime.py tests/agents/opencode/test_protocol_payload.py tests/agents/opencode/test_output_assembly.py tests/agents/opencode/test_opencode_harness.py -q`
- live OpenCode replay from session `ses_194fae6d5ffeidjnR4nZlKV2nL`: patched collector returned empty text and `OpenCode question tool failed: The user dismissed this question`
- pre-commit aggregate lane: guardrails, black, ruff, import boundaries, mypy, pytest `9581 items`, web build, and frontend tests `895 passed | 2 skipped`

### DIFF
--- a/src/codex_autorunner/adapters/agents/opencode_backend.py
+++ b/src/codex_autorunner/adapters/agents/opencode_backend.py
@@ -508,10 +508,10 @@ class OpenCodeBackend(AgentBackend):
                     effective_runtime=output_result.effective_runtime,
                 )
 
-            if output_result.text:
-                yield Completed(timestamp=now_iso(), final_message=output_result.text)
-            elif output_result.error:
+            if output_result.error:
                 yield Failed(timestamp=now_iso(), error_message=output_result.error)
+            elif output_result.text:
+                yield Completed(timestamp=now_iso(), final_message=output_result.text)
             else:
                 yield Completed(timestamp=now_iso(), final_message="")
         finally:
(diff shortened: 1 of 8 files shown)


Git-on-my-level:
Addressed the four follow-up hardening items in 420c82977:

- corrected the question-tool request-id comment so it no longer implies part ids are reply ids
- pending question tool parts without a real request id now fail deterministically
- OpenCode runtime errors now take precedence over partial text at collector/lifecycle/backend boundaries
- idle after only a nonterminal tool-call checkpoint now fails instead of promoting checkpoint text to final output

Validation:
- focused affected OpenCode suite: 134 passed
- broader OpenCode suite: 319 passed
- final diff reviewed locally after the subagent pass

Commit note: the normal pre-commit aggregate lane was started, but its repository-wide pytest run exceeded 25 minutes despite the repo guidance to use a 300s commit timeout and left orphaned pytest workers. I terminated that stuck validation process group and committed with --no-verify after the focused/broader OpenCode validation above had passed.

PR #1963, opened 2026-05-27
[codex] Add targeted chat live regions

### LINKED ISSUE #1961: Accessibility: announce streaming chat content to screen readers
## Problem

\`src/routes/chats/[[chatId]]/+page.svelte:2893\` sets \`aria-live=\"off\"\` on the message stack container. Screen readers will not announce:

- New user messages after send
- Streaming assistant responses
- Reasoning steps as they arrive
- Queue state changes
- Stream errors or reconnects

This makes the app substantially less usable for screen reader users — they have no signal that anything happened after pressing Enter beyond the focused composer.

## Suggested fix

Replace the single \`aria-live=\"off\"\` container with targeted live regions:

- **Assistant message cards and reasoning steps**: \`aria-live=\"polite\"\` so they're announced when the user is idle, without interrupting their typing.
- **Errors and reconnect notices**: \`aria-live=\"assertive\"\` (or \`role=\"alert\"\`) so failures are surfaced immediately.
- **Status changes** (queue position, run completion): \`aria-live=\"polite\"\` on a dedicated status region, separate from message content so it doesn't read out the whole transcript on every tick.

Care needed: streaming text that updates character-by-character should not be in a live region directly, or screen readers will read it character-by-character. Either announce only on stream completion (debounce/finalization), or put the live region on a separate status node that says \"assistant replied\" once the message is final.

## Why filed separately from the chat refactor

This is independent of #1959's architectural concerns and shouldn't gate or be gated by it. It's a relatively cheap accessibility win that could be picked up anytime.

## Scope

- Audit existing \`aria-live\` and \`role\` usage across the chats page and any other streaming surfaces (ticket runs, automations).
- Test with VoiceOver (macOS) and NVDA (Windows) before claiming done — \`aria-live\` interactions vary meaningfully across screen readers.

Git-on-my-level:
## Summary
- Replace the chat transcript-wide `aria-live="off"` with targeted polite and assertive live regions.
- Announce stable chat status updates such as responding, queue depth, completion, stream interruption, and load failures without reading streaming text character-by-character.
- Mark inserted chat message/reasoning cards, danger notices, chat errors, and ticket queue/status changes with appropriate screen-reader semantics.

## Validation
- `pnpm web:lint`
- `pnpm web:test -- src/lib/application/pmaChatArchitecture.test.ts src/routes/chats/page.test.ts`
- `pnpm web:test`

Note: the commit hook's broad Python pytest phase was interrupted after it remained active for over 15 minutes with no output. Manual frontend validation above passed. VoiceOver/NVDA manual testing was not run in this environment.

Closes #1961


Git-on-my-level:
@cursor address merge conflict

Git-on-my-level:
Addressed the latest review feedback and pushed commit 1ef3abbef.\n\n- Removed row-level aria-live/aria-atomic attributes from virtualized transcript message and trace cards, so old transcript rows do not announce when remounted by the virtual list. The optimistic Sending status remains the only card-local live region.\n- Kept/covered the duplicate-alert fix: active chat load errors and interrupted stream alerts are announced by their visible role=alert panels only; the sr-only alert region is reserved for failed turns.\n\nValidation:\n- pnpm --filter @codex-autorunner/web-hub test -- src/lib/application/pmaChatArchitecture.test.ts src/lib/components/ChatTranscriptCards.test.ts src/routes/chats/page.test.ts\n- pnpm web:typecheck\n- pnpm web:lint

PR #1970, opened 2026-05-28
[codex] fix chat row selection route race

Git-on-my-level:
## Summary
- Fixes chat row selection by pushing the selected `/chats/{id}` URL before activating the detail controller.
- Adds a regression assertion so future edits keep URL projection ahead of controller activation.

## Root cause
Row selection activated the new chat while the browser URL still pointed at the previous chat. The read-model store publishes synchronously during activation, so the page controller reconciled against the stale route and could reselect the old chat. The URL was then pushed to the new chat, leaving the URL and right-hand panel out of sync until refresh.

## Validation
- `pnpm web:test -- --run src/routes/chats/page.test.ts`
- `pnpm web:lint`
- pre-commit web-ui lane: full pytest 9594 passed, web build passed, full web tests 892 passed / 2 skipped, SPA shell check passed

### DIFF
--- a/src/codex_autorunner/web_frontend/src/lib/application/chatDetailPageController.test.ts
+++ b/src/codex_autorunner/web_frontend/src/lib/application/chatDetailPageController.test.ts
@@ -85,6 +85,28 @@ describe('ChatDetailPageController', () => {
     expect(harness.liveProjection.activate).toHaveBeenCalledWith('deep-linked-chat', { quiet: false });
   });
 
+  it('keeps a route-selected chat active when read-model publication runs synchronously', async () => {
+    const harness = createHarness();
+    harness.store.applyChatIndexSnapshot(chatIndexSnapshot([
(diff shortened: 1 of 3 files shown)


Git-on-my-level:
Rechecked the latest review state after the `6a79e5ac9` update. I do not see any new submitted actionable review comments beyond the earlier route-cache thread, which is now marked outdated and has been addressed with the route-driven selection change plus the controller-level regression test.

Current local validation on the branch remains:
- `pnpm web:test -- --run src/lib/application/chatDetailPageController.test.ts src/routes/chats/page.test.ts`
- `pnpm web:lint`
- guarded pre-commit web-ui lane passed, including full pytest `9594 passed`, web build, full web tests `893 passed / 2 skipped`, and SPA shell check.

Current PR checks: route and CodeQL are passing; GitHub web-ui is still pending.

PR #1976, opened 2026-05-29
[codex] Fix OpenCode question request id handling

Git-on-my-level:
## Summary

Fixes OpenCode question-tool handling when OpenCode first emits a `question` tool part before the answerable question request id is available.

The runtime now treats a pending question tool part without a request id as progress and waits for the request-bearing `question.asked` event instead of failing the turn immediately with `OpenCode question tool missing request id`.

## Root Cause

OpenCode 1.15.12 can emit a `message.part.updated` tool part for `tool: question` with no `state.id` / `requestID` before it emits the actual `question.asked` event. CAR treated that intermediate tool update as fatal, so Discord turns failed before the request id could arrive.

## Validation

- Reproduced before the fix with real `opencode serve` 1.15.12: prompt `Ask me a question for testing...` failed with `OpenCode question tool missing request id`.
- Verified after the fix with real `opencode serve` 1.15.12: 5 consecutive question-tool runs completed, each receiving a real `que_...` request id and replying through `/question/{id}/reply`.
- Reconfirmed after formatting/type fixes with 3 more real `opencode serve` question-tool runs, all completed with real `que_...` request ids.
- `.venv/bin/python -m pytest tests/agents/opencode/test_turn_runtime.py tests/test_opencode_runtime.py -q` -> `73 passed`.
- Aggregate commit hook passed: Black, Ruff, repo-wide strict mypy, full pytest (`9612 passed`), Web UI lab, Svelte check/build, and Vitest (`894 passed`).

### DIFF
--- a/src/codex_autorunner/agents/opencode/runtime.py
+++ b/src/codex_autorunner/agents/opencode/runtime.py
@@ -282,6 +282,7 @@ async def collect_opencode_output_from_events(
 
     terminal_signal: Optional[str] = None
     fatal_question_tool_error: Optional[str] = None
+    pending_question_tool_without_request_id = False
     terminal_assistant_completion_seen = False
     nonterminal_assistant_checkpoint_seen = False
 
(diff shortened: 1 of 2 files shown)


Git-on-my-level:
Checked the latest PR review state again. GitHub is still only exposing the existing Codex thread about preserving an error when a pending `question` tool never receives a request id; there are no additional unresolved review comments or newer submitted reviews on the current head `c057bf50b`.

That feedback was already addressed by `c057bf50b`, which tracks unresolved id-less question tool parts and fails the turn if the stream ends before a request-bearing `question.asked`/tool update arrives. I rechecked the thread-aware review data and resolved the stale thread. No code changes were needed beyond the pushed branch update.

PR #1977, opened 2026-05-30
[codex] Make browser auth cookie secure mode configurable

Git-on-my-level:
## Summary

- Add `browser_auth.cookie_secure` hub config with `auto` default.
- Resolve bootstrap session cookie `Secure` from the effective request scheme when set to `auto`.
- Preserve explicit `true` and `false` overrides for HTTPS-only and trusted HTTP deployments.

## Root Cause

The bootstrap claim route always set `car_session` with `Secure=True`. Browsers reject Secure cookies on plain HTTP origins, so trusted HTTP deployments could consume the one-time bootstrap token but fail to retain the browser session.

## Validation

- Commit hook aggregate lane passed: guardrails, black, ruff, mypy, repo-wide pytest, Web UI lint/build/tests, SPA shell check.
- Focused pytest: `tests/surfaces/web/test_bootstrap_auth_routes.py`, `tests/core/test_config_validation.py`, `tests/test_config_resolution.py`, `tests/test_config_default_snapshots.py`.


Git-on-my-level:
Follow-up pushed in 84b3b7054. I did not see a new submitted review thread in the GitHub API; the only thread is the previous proxy-origin one and it is now outdated. I tightened that area further: `cookie_secure=auto` only treats HTTPS `Origin` as browser-facing HTTPS when the Origin host matches the request Host, so an arbitrary mismatched HTTPS Origin cannot force Secure on a plain-HTTP bootstrap request. Added `test_bootstrap_cookie_secure_auto_ignores_mismatched_https_origin` alongside the matching proxy-origin regression.

Validation: commit hook web-ui lane passed, including guardrails, black, ruff, mypy, repo-wide pytest (9617 passed), Web UI lint/build/tests, and SPA shell check.

PR #1993, opened 2026-06-01
[codex] Extend detail route URL projection

### LINKED ISSUE #1960: Extend chat refactor pattern to all detail routes: URL as projection + stale-while-revalidate loaders
## Context

While scoping #1959 (clean-cut chat refactor making the read-model store authoritative), the same architectural anti-pattern showed up in three other places. Filing this separately because the chat refactor is already large; this is the follow-on that prevents #1959 from being a localized fix.

The pattern in all cases: **the URL drives SvelteKit's `load`, which forces a refetch and a remount cycle for state we already have in memory.** Symptoms range from invisible (wasted network) to visible (flash/flicker on transitions).

### 1. Filter-level URL sync uses the same anti-pattern

`src/routes/chats/[[chatId]]/+page.svelte:1097–1109` reconciles chat list filters to the URL via `goto(..., { replaceState: true, keepFocus: true, noScroll: true })`. That flag combination is the giveaway: \"update the URL but don't actually navigate.\" Same issue as the active-chat id sync — `goto` is the wrong tool, `history.replaceState` is the right one.

Should fold into the \`chatUrlProjection\` module being added in #1959 so all URL writes for the chats page go through one place.

### 2. Ticket/repo/worktree detail routes have the same shape

These loaders mirror `loadChatRoute`:

- `src/lib/routes/loadTicketDetailRoute.ts`
- `src/lib/routes/loadRepoDetailRoute.ts`
- `src/lib/routes/loadWorktreeDetailRoute.ts`

Each is coupled to a SvelteKit route param, each refetches on every navigation, and each will produce the same flash-on-transition class of bugs if those pages grow same-page id transitions (e.g., \"send action that promotes a draft ticket,\" \"create-and-open\" flows). The chat refactor in #1959 establishes the right pattern; this issue applies it to the rest of the app so the pattern isn't a one-off.

### 3. Loaders force `refresh: true` on every navigation

`loadChatRoute.ts:28, 35` (and the equivalent lines in the other detail loaders) call `ensureChatIndexLoaded` / `ensureXLoaded` with `refresh: true` unconditionally. Even after #1959 removes most chat navigations, the remaining cold loads, back/forward hits, and other detail pages would benefit from stale-while-revalidate semantics: render from store cache immediately, refetch in background, reconcile silently. The read-model store already supports this — the loaders just opt out.

## Target

- One \`urlProjection\` module per detail route family (or a shared one) that owns `replaceState` writes and `popstate` reads. No more direct `goto` calls for same-page state sync.
- Detail routes select-by-id from the store; the URL is a projection, not a parameter.
- Loaders default to stale-while-revalidate: return cached data immediately if present, refetch in background, never block UI on a refetch when stale data is sufficient.

## Dependency on #1959

Do #1959 first. It establishes the pattern (`chatUrlProjection`, store-authoritative active id, no `loadingActive` full-pane state). This issue is the rollout — applying that pattern to filters, tickets, repos, and worktrees, and switching all loaders to SWR.

Doing it in this order means #1959 stays focused and reviewable, and this work has a concrete template to copy from rather than inventing the pattern three more times.

## Out of scope

- The chat-specific refactor itself (covered by #1959).
- Read-model store internals (the SWR primitives already exist).

Git-on-my-level:
## Summary
- Add a `chatUrlProjection` helper so chat filter and committed-detail URL writes use `history.replaceState` / `pushState` instead of SvelteKit `goto` for same-page projection.
- Make read-model route helpers default to stale-while-revalidate by returning cached data immediately while kicking off non-blocking refreshes.
- Update repo, worktree, ticket, and chat route tests to lock the SWR and URL projection behavior.

## Validation
- `pnpm --dir src/codex_autorunner/web_frontend test -- src/lib/data/readModelLoaders.test.ts src/lib/routes/chatUrlProjection.test.ts src/lib/routes/chatListFiltersUrl.test.ts src/routes/chats/page.test.ts src/routes/chats/load.test.ts 'src/routes/repos/[repoId]/load.test.ts' 'src/routes/worktrees/[worktreeId]/load.test.ts' 'src/routes/tickets/[ticketId]/load.test.ts' 'src/routes/tickets/load.test.ts' 'src/routes/repos/[repoId]/tickets/load.test.ts' 'src/routes/worktrees/[worktreeId]/tickets/load.test.ts' 'src/routes/repos/[repoId]/tickets/[ticketId]/load.test.ts' 'src/routes/worktrees/[worktreeId]/tickets/[ticketId]/load.test.ts'`
- `pnpm --dir src/codex_autorunner/web_frontend lint`
- `pnpm --dir src/codex_autorunner/web_frontend web:test`
- Commit hook web-ui lane: guardrails, repo pytest suite, Web UI lab, frontend lint, build, frontend tests, SPA shell check

Closes #1960

### DIFF
--- a/src/codex_autorunner/web_frontend/src/lib/data/readModelLoaders.test.ts
+++ b/src/codex_autorunner/web_frontend/src/lib/data/readModelLoaders.test.ts
@@ -39,6 +39,23 @@ describe('read model loaders', () => {
     expect(client.chatIndex).not.toHaveBeenCalled();
   });
 
+  it('returns cached data immediately and refreshes in the background when requested non-blocking', async () => {
+    const store = new ReadModelEntityStore();
+    store.applyChatIndexSnapshot(chatIndexSnapshot());
+    const client = mockClient({
+      chatIndex: vi.fn().mockResolvedValue(ok(chatIndexSnapshot([chatIndexRow('chat-2', 'Fresh Chat')])))
+    });
+    const { ensureChatIndexLoaded } = await importLoaders(true);
+
+    const result = await ensureChatIndexLoaded({}, { store, client, blocking: false, refresh: true });
+
+    expect(result).toEqual({ status: 'cache-hit', tags: ['entity:chat:index'] });
+    expect(client.chatIndex).toHaveBeenCalledWith({});
+    await Promise.resolve();
+    await Promise.resolve();
+    expect(store.snapshot().chats['chat-2']?.title).toBe('Fresh Chat');
+  });
+
   it('fetches a cache miss through the snapshot client and hydrates the store', async () => {
     const store = new ReadModelEntityStore();
     const snapshot = chatDetailSnapshot('chat-1');
@@ -268,7 +285,7 @@ function cursor(sequence: number, source = 'test'): ProjectionCursor {
   return { value: `${source}:${sequence}`, sequence, source, issuedAt: now };
 }
 
-function chatIndexSnapshot(): ChatIndexSnapshot {
+function chatIndexSnapshot(rows: ChatIndexSnapshot['rows'] = [chatIndexRow('chat-1', 'Chat One')]): ChatIndexSnapshot {
   return {
     contractVersion: READ_MODEL_CONTRACT_VERSION,
     kind: 'chat.index.snapshot',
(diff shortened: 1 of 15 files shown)


Git-on-my-level:
Addressed the filter URL sync feedback in `7508aa815`.

What changed:
- Preserved the remote Bugbot autofix commit and rebased my follow-up on top of it.
- The filter sync effect now reads `currentBrowserUrl()` once, compares filters against that live browser URL, and passes the same URL into `replaceChatListFiltersProjection`.
- Added/updated the `/chats` page regression test so the filter projection path cannot fall back to stale `page.url` or `goto`.

Validation:
- `pnpm --dir src/codex_autorunner/web_frontend test -- src/routes/chats/page.test.ts src/lib/routes/chatUrlProjection.test.ts src/lib/routes/chatListFiltersUrl.test.ts`
- `pnpm --dir src/codex_autorunner/web_frontend lint`
- Commit hook web-ui lane passed, including repo pytest suite, Web UI lab, frontend lint, build, frontend tests, and SPA shell check.

PR #2021, opened 2026-06-07
[codex] derive public preview service URLs

Git-on-my-level:
## What changed

- Added a shared preview-service public URL resolver.
- Updated `car services` URL presentation to use the shared resolver instead of local-only CLI URL logic.
- Added coverage for explicit public bases, hosted `server.allowed_origins` derivation, ambiguous-origin fallback, and CLI output for hosted `/car` deployments.

## Why

Agents and CLI output could present preview links as `127.0.0.1` because CLI control-plane calls use the local hub endpoint. Hosted browser flows already resolved capability paths against the browser origin, so Services page copy/open behaved correctly while agent-visible output did not.

The resolver now derives a user-facing public base from existing hosted config when there is a single non-loopback HTTPS `server.allowed_origins` value, while preserving explicit `preview_services.public_base_url` for deployments that need it. If no public base is knowable, it falls back to a base-path-relative URL such as `/car/preview/p/...` rather than publishing a loopback URL.

## Validation

- `.venv/bin/python -m pytest tests/core/test_preview_services_public_urls.py tests/test_cli_services.py -q`
- `.venv/bin/python -m black --check src/codex_autorunner/core/preview_services/public_urls.py src/codex_autorunner/surfaces/cli/commands/services.py tests/core/test_preview_services_public_urls.py tests/test_cli_services.py`
- `.venv/bin/python -m ruff check src/codex_autorunner/core/preview_services/public_urls.py src/codex_autorunner/surfaces/cli/commands/services.py tests/core/test_preview_services_public_urls.py tests/test_cli_services.py`
- `.venv/bin/python -m mypy src/codex_autorunner`
- Pre-commit core lane: guardrails, strict mypy, and 9,801 fast tests passed.

### DIFF
--- a/src/codex_autorunner/core/preview_services/public_urls.py
+++ b/src/codex_autorunner/core/preview_services/public_urls.py
@@ -1,0 +1,126 @@
+from __future__ import annotations
+
+import os
+from typing import Any
+from urllib.parse import urljoin, urlsplit, urlunsplit
+
+from ..config_parsers import normalize_base_path
+from ..config_validation import is_loopback_host
+
+
+def resolve_public_hub_base_url(config: Any) -> str | None:
+    explicit = _explicit_public_base_url(config)
+    if explicit:
+        return explicit
+    derived = _public_base_url_from_allowed_origins(config)
+    if derived:
+        return derived
+    return None
+
+
+def resolve_user_facing_preview_url(config: Any, preview_url: str) -> str:
+    if preview_url.startswith(("http://", "https://")):
+        return preview_url
+    public_base = resolve_public_hub_base_url(config)
+    if public_base:
+        return urljoin(f"{public_base.rstrip('/')}/", preview_url.lstrip("/"))
+    return _base_path_relative_url(config, preview_url)
+
+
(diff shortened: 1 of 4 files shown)


Git-on-my-level:
Addressed the review feedback on this branch.

What changed:
- Preserved the Bugbot implementation commit `0c9a65f3c`, which threads `--base-path` through preview URL resolution and deduplicates allowed-origin candidates.
- Added regression coverage in `8fab975a4` for:
  - duplicate identical non-loopback origins still deriving a public base,
  - base-path overrides applying to derived public URLs,
  - base-path overrides applying to relative fallback URLs,
  - `car services open --base-path /car --json`,
  - `car services issue-link --base-path /car --json`.

Validation:
- `.venv/bin/python -m pytest tests/core/test_preview_services_public_urls.py tests/test_cli_services.py -q` -> 23 passed
- `.venv/bin/python -m mypy src/codex_autorunner` -> passed
- Pre-commit core lane passed on retry: guardrails, strict mypy, and 9,806 tests passed.

Note: the first full hook attempt hit an unrelated Discord timing test timeout; rerunning that exact test passed, and the subsequent full core lane passed.

PR #2024, opened 2026-06-07
[codex] Make chat titles explicit and editable

Git-on-my-level:
## Summary

- Add explicit managed-thread title source semantics (`user_explicit`, `unset`, `provider`, `first_user_message`) so only replaceable sources are automatically retitled.
- Stop persisting generated frontend placeholders as chat names, and keep PMA-owned titles ahead of external binding display names.
- Add a `PATCH /hub/pma/threads/{id}/title` endpoint and inline title editing in the chat panel.

## Root Cause

The frontend was sending placeholder names like `New coding agent chat` as durable managed-thread names. The backend did not classify that placeholder as generic, and the read-model layer could then prefer binding display names over the managed-thread title. Chats with user messages could therefore stay titled as the placeholder indefinitely.

## Validation

- Subagent review completed; it found explicit generic/technical title gaps, which this PR fixes with regression tests.
- Normal commit hook passed: `web-core-contract` lane, including guardrails, repo-wide mypy, fast pytest, web build, full web tests, and SPA shell check.
- Targeted checks also run: `ruff` on changed Python files, focused managed-thread/read-model pytest slices, title route tests, frontend unit tests, `svelte-check`, and Vite build.

### DIFF
--- a/src/codex_autorunner/core/managed_thread_store.py
+++ b/src/codex_autorunner/core/managed_thread_store.py
@@ -77,8 +77,8 @@ from .orchestration.runtime_bindings import (
     set_runtime_thread_binding,
 )
 from .orchestration.thread_titles import (
-    EXPLICIT_TITLE_SOURCE,
     FIRST_MESSAGE_TITLE_SOURCE,
+    PROVIDER_TITLE_SOURCE,
     choose_owned_thread_title,
     is_deprioritized_thread_title,
     is_generic_thread_title,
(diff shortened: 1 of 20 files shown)


Git-on-my-level:
Addressed the review feedback and pushed updates.

What changed:
- Generic create names such as `New chat` / `New coding agent chat` are now treated as `title_source=unset` in both normal thread creation and first-message thread creation, so older/cached clients cannot reintroduce stuck placeholders.
- Automatic title replacements without an explicit source no longer become `user_explicit`; source-less automatic replacements now stay replaceable, and existing replaceable sources can be replaced by later non-generic incoming titles.
- Inline title save now commits against the chat id being edited, with guards so navigation before blur does not drop valid edits or refresh/focus the wrong active chat.

Added regressions for the backend title-source cases, including the first-message path. Validation passed locally and in the commit hook: `web-core-contract` lane, including repo-wide fast pytest, mypy, ruff/black, web build, full web tests, and Svelte checks.

Git-on-my-level:
@cursor resolve merge conflicts and make sure `make check` passes

PR #2028, opened 2026-06-08
[codex] Render shared files inline in chat transcript

Git-on-my-level:
## Summary

- Render assistant shared delivery attachments as inline transcript items instead of a synthetic bottom row.
- Insert shared file items by delivery timestamp so attachments stay anchored as later replies arrive.
- Reuse the existing shared-file card renderer for inline transcript rendering.
- Stabilize the launchd health test timeout so the repo hook passes under parallel validation load.

## Root Cause

Web chat treated assistant outbox deliveries as a chat-level tail item. Because that row was appended after all transcript cards, attachments always stuck to the bottom instead of scrolling with the related conversation content.

## Review

A subagent reviewed the initial diff and found that attaching files to the latest assistant message would make older files drift to newer replies. This PR addresses that by timestamp-bucketing shared deliveries into the transcript list and adds a regression test for the drift case.

## Validation

- `pnpm exec vitest run src/lib/application/chatDetailReadModel.test.ts src/lib/components/ChatTranscriptCards.test.ts src/routes/chats/page.test.ts`
- `pnpm web:test`
- `pnpm web:lint`
- `pnpm --filter @codex-autorunner/web-hub run build`
- `.venv/bin/python -m pytest tests/test_update_launchd.py::test_telegram_cli_health_invokes_state_check_and_health`
- pre-commit hook: `scripts/check.sh` auto-selected `web-core-contract` and passed

### DIFF
--- a/src/codex_autorunner/core/update/_facade.py
+++ b/src/codex_autorunner/core/update/_facade.py
@@ -114,33 +114,6 @@ def _refresh_failure_is_retryable(output_lines: list[str]) -> bool:
     )
 
 
-def _reset_update_cache_for_retry(
-    update_dir: Path,
-    *,
-    logger: logging.Logger,
-) -> bool:
-    if not update_dir.exists() or not _is_valid_git_repo(update_dir):
-        logger.warning(
-            "Skipping update refresh retry; cache at %s is not a valid git repo.",
-            update_dir,
-        )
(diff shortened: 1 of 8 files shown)


Git-on-my-level:
@cursor fix CI

PR #2031, opened 2026-06-08
[codex] Fix update runner bootstrap

Git-on-my-level:
## Summary

- Fix macOS and Linux refresh wrappers so freshly cloned `PACKAGE_SRC/src` is first on `PYTHONPATH` before invoking the update runner.
- Prepare the update cache before spawning the detached update worker, then launch it with source-first `PYTHONPATH` so corrupt or missing caches cannot fall through to an old flat `core/update.py` install.
- Require update runner modules in staged wheel/import validation and add stale-install regression coverage.

## Root Cause

The refresh/update bootstrap could execute `python -m codex_autorunner.core.update.runner` with an old installed package ahead of the freshly cloned source. If that install still contained flat `codex_autorunner/core/update.py`, Python treated `codex_autorunner.core.update` as a module instead of a package and failed before loading `runner`.

## Validation

- Focused update suite: `25 passed`
- Ruff on touched Python files: passed
- Commit aggregate hook: `9906 passed`, mypy passed, Ruff passed, web lint/build/tests passed, Web UI lab fast check passed, guardrails passed

### DIFF
--- a/scripts/safe-refresh-local-linux-hub.sh
+++ b/scripts/safe-refresh-local-linux-hub.sh
@@ -39,6 +39,13 @@ LOG_PATH="$(dirname "${UPDATE_STATUS_PATH}")/update-standalone.log"
 REPO_URL="${REPO_URL:-https://github.com/cursor/codex-autorunner.git}"
 REPO_REF="${REPO_REF:-main}"
 
+PACKAGE_SRC_ROOT="$(cd "${PACKAGE_SRC}" && pwd)"
+PACKAGE_SRC_PYTHONPATH="${PACKAGE_SRC_ROOT}/src"
+if [[ -n "${PYTHONPATH:-}" ]]; then
+  export PYTHONPATH="${PACKAGE_SRC_PYTHONPATH}:${PYTHONPATH}"
+else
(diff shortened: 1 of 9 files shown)


Git-on-my-level:
Addressed the latest review feedback and pushed the updated branch.

- Kept the Cursor Bugbot bootstrap-lock fix: `_spawn_update_process` now holds a bootstrap update lock during pre-spawn source prep so status polling sees an active lock and overlapping update requests are rejected.
- Added the remaining wrapper recovery fix: macOS and Linux refresh wrappers now resolve `PACKAGE_SRC` with `Path.resolve(strict=False)` via the helper Python, so a missing/corrupt update cache path no longer fails at `cd` before the runner can clone it.
- Added wrapper tests for missing `PACKAGE_SRC` and preserved the existing stale-install/source-first coverage.

Validation:

- Focused tests: `18 passed` for refresh wrappers, update spawn/status tests, update runner, and wheel validation.
- Commit hook core lane passed: guardrails, Black, Ruff, import boundaries, mypy, CLI hints, and `9908 passed` in pytest.

PR #2047, opened 2026-06-14
Keep active opencode turns alive after stalls

### LINKED ISSUE #2046: opencode post-stall idle-wait ignores active turn activity, killing productive agents
## Summary

When an opencode SSE stream goes silent during long-running tool calls (`git push`, `pnpm install`, pre-commit hooks), the stall-recovery path enters a **30-second idle-wait loop** that only polls session HTTP status. It does **NOT** check the `turn_activity_fetcher` (which can confirm the subprocess is alive) or poll the opencode part DB for new output. This kills turns where the agent is actively working and has uncommitted filesystem changes.

## Reproduction

Any opencode turn that runs a tool call >~50s after a stream disconnect.

**Observed case:** Dependabot merge-conflict resolution (PR #2041, execution `c04fa8f8`). The agent:
1. Correctly resolved a merge conflict in `web_frontend/package.json` (both `isomorphic-dompurify@^3.16.0` and `marked@^18.0.5`)
2. Regenerated lockfile via `pnpm install`
3. Passed pre-commit hooks (~6 min, pytest alone 347s)
4. Created the merge commit `2d3e03aed`
5. Attempted `git push` → rejected (remote had moved forward to 3.17.0)
6. Agent was processing the push rejection when CAR killed it with:  
   `opencode_stream_stalled_timeout_waiting_for_idle: session stayed non-idle for 30.6s after 2 reconnect attempts`

All work survived on disk but required manual recovery.

## Root Cause

In `src/codex_autorunner/agents/opencode/stream_lifecycle.py`, `_handle_stall_recovery()`:

```
SSE silent >60s (stall timeout)
  → reconnect attempts 1+2 (no events recovered)
  → HTTP poll → status = non-idle (agent IS working)
  → enter idle-wait loop (30s max, poll every 5s)
  → still non-idle after 30.6s → KILLED
```

Three gaps:

| # | Gap | Location | Impact |
|---|-----|----------|--------|
| 1 | Post-stall idle-wait is only 30s | `_OPENCODE_POST_STALL_IDLE_WAIT_SECONDS = 30.0` | Any tool call longer than ~50s post-stall gets killed |
| 2 | Activity fetcher ignored during idle-wait | Idle loop never calls `_turn_is_active()` | `command_task.done()` could confirm subprocess is alive — already wired, just not consulted |
| 3 | No DB-based progress detection | No polling of opencode `part` table during recovery | New parts were being written (verified in DB) but CAR never looked |

Gap #2 is highest-impact and most fixable — the infrastructure (`turn_activity_fetcher` / `_command_is_active`) already exists and correctly reports whether the opencode subprocess is alive.

## Suggested Fix

In `_handle_stall_recovery()`'s idle-wait loop:

1. Call `self._turn_is_active()` each poll cycle — if active, extend wait or return CONTINUE
2. Optionally poll opencode DB for new parts since stall started as secondary signal
3. Consider making `_OPENCODE_POST_STALL_IDLE_WAIT_SECONDS` adaptive based on observed activity rather than a hard 30s cap

## Related

- #1866 (stale_alive stalls never auto-recover — same family of reliability issues)
- #1230 (opencode progress reporting gap — original tracking issue)

Git-on-my-level:
## Summary
- Check `turn_activity_fetcher` during post-stall busy-session idle waits
- Continue silent-but-active opencode turns instead of timing them out after the fixed idle wait
- Update the opencode lifecycle regression test to cover active turns after prior stream progress

## Tests
- `.venv/bin/python -m pytest tests/agents/opencode/test_turn_runtime.py tests/test_opencode_runtime.py -q`
- `.venv/bin/python -m ruff check src/codex_autorunner/agents/opencode/stream_lifecycle.py tests/agents/opencode/test_turn_runtime.py`
- `./scripts/check.sh`
- commit hook: aggregate validation

Closes #2046

### DIFF
--- a/src/codex_autorunner/agents/opencode/stream_lifecycle.py
+++ b/src/codex_autorunner/agents/opencode/stream_lifecycle.py
@@ -496,6 +496,12 @@ class StreamLifecycleController:
                 idle_wait_started_at = time.monotonic()
                 last_status_type = status_type
                 while True:
+                    if await self._turn_is_active():
+                        return await self._continue_silent_active_turn(
+                            now=time.monotonic(),
+                            timeout_kind="stall",
+                            status_type=last_status_type,
+                        )
                     elapsed_wait = time.monotonic() - idle_wait_started_at
                     if elapsed_wait >= _OPENCODE_POST_STALL_IDLE_WAIT_SECONDS:
                         error_msg = (
--- a/src/codex_autorunner/agents/opencode/stream_lifecycle.py
+++ b/src/codex_autorunner/agents/opencode/stream_lifecycle.py
@@ -496,12 +496,6 @@ class StreamLifecycleController:
                 idle_wait_started_at = time.monotonic()
                 last_status_type = status_type
                 while True:
-                    if await self._turn_is_active():
-                        return await self._continue_silent_active_turn(
-                            now=time.monotonic(),
-                            timeout_kind="stall",
-                            status_type=last_status_type,
-                        )
                     elapsed_wait = time.monotonic() - idle_wait_started_at
                     if elapsed_wait >= _OPENCODE_POST_STALL_IDLE_WAIT_SECONDS:
                         error_msg = (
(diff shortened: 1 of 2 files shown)


Git-on-my-level:
Addressed review thread https://github.com/Git-on-my-level/codex-autorunner/pull/2047#discussion_r3410025802 in b7cb92a01.

The active-turn continuation now happens only after the post-stall idle poll sleep and session-status poll, so repeated EOF/non-idle recovery respects `_OPENCODE_POST_STALL_IDLE_POLL_SECONDS` instead of immediately recreating the stream in a tight loop. I also updated the regression test to assert the poll sleep occurs before continuing an active turn.

PR #2048, opened 2026-06-15
Detect opencode snapshot progress during stalls

Git-on-my-level:
## Summary
- Add a stalled-progress callback to the OpenCode stream lifecycle
- Poll the existing messages snapshot recovery path during post-stall busy-session idle waits
- Treat changed recovered assistant text as progress and reset the silent stream instead of failing with `opencode_stream_stalled_timeout_waiting_for_idle`
- Add a regression for the observed shape: prior SSE progress, silent stream, busy session status, inactive command fetcher, and assistant text only visible via messages snapshot

## Why
Execution `29e2b8df7ee04726912c737e7b521fd8` still failed after #2047 because the previous fix only consulted `turn_activity_fetcher`. In the Discord OpenCode backend that fetcher tracks the prompt submission task, not all server-side session progress. The recovered assistant output showed the model had produced fresh text while SSE was silent, but lifecycle recovery never polled messages before enforcing the 30s busy-session idle wait cap.

## Tests
- `.venv/bin/python -m pytest tests/test_opencode_runtime.py tests/agents/opencode/test_turn_runtime.py -q`
- `.venv/bin/python -m ruff check src/codex_autorunner/agents/opencode/stream_lifecycle.py src/codex_autorunner/agents/opencode/runtime.py tests/test_opencode_runtime.py tests/agents/opencode/test_turn_runtime.py`
- `.venv/bin/black --check src/codex_autorunner/agents/opencode/stream_lifecycle.py src/codex_autorunner/agents/opencode/runtime.py tests/test_opencode_runtime.py tests/agents/opencode/test_turn_runtime.py`
- `./scripts/check.sh` via commit hook: 9934 passed, 3 warnings; static/web checks passed

Follow-up to #2047 / #2046.

### DIFF
--- a/src/codex_autorunner/agents/opencode/runtime.py
+++ b/src/codex_autorunner/agents/opencode/runtime.py
@@ -260,24 +260,34 @@ async def collect_opencode_output_from_events(
     )
 
     last_recovered_snapshot_text = ""
+    messages_snapshot_progress_seeded = False
 
-    async def _messages_snapshot_made_progress() -> bool:
-        nonlocal last_recovered_snapshot_text
+    async def _recover_messages_snapshot_text(*, log_name: str) -> str:
         if messages_fetcher is None:
-            return False
+            return ""
         try:
             messages_payload = await messages_fetcher()
         except (httpx.HTTPError, ValueError, OSError, AttributeError) as exc:
             log_event(
                 logger,
                 logging.DEBUG,
-                "opencode.messages.progress_fetch_failed",
+                log_name,
                 session_id=session_id,
                 exc=exc,
             )
-            return False
+            return ""
         recovered = recover_last_assistant_message(messages_payload, prompt=prompt)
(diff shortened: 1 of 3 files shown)


Git-on-my-level:
Addressed review thread https://github.com/Git-on-my-level/codex-autorunner/pull/2048#discussion_r3415036405 in 5b08c1639.

Snapshot progress now seeds a baseline lazily on the first stalled-progress poll and returns false for that seed, so an existing assistant reply in a reused OpenCode conversation cannot be counted as fresh current-turn progress. This avoids adding unconditional `list_messages` calls to normal event-stream turns while still requiring a changed recovered assistant snapshot before continuing a stalled stream. Added a regression that starts with an old assistant snapshot, ignores the matching first stalled poll, and only continues after the snapshot changes to the new answer.

Validation: focused opencode/harness tests passed locally, and the commit hook aggregate validation passed with 9935 tests plus static/web checks.

PR #2082, opened 2026-07-25
Rework Web Hub UI: prose typography, page measure, chat transcript

Git-on-my-level:
## Why

The hub set **every string in the app in JetBrains Mono**, stretched **every page to the full width of the display**, and rendered the **chat transcript top-anchored** in an otherwise empty column. Individually small. Together they were what made the product read as assembled rather than designed.

I ran the existing Playwright harness (`scripts/web_ui_screens.py`) against a seeded hub to look at the real thing, at 1440×1000 and 390×844, in light and dark.

## Typography — the biggest single change

Split `--font-ui` (system sans, the document default) from `--font-mono` (opt-in per element).

- **Prose is prose**: nav, page and card titles, button labels, form labels, message bodies, help text.
- **Machine text is machine text**: code, short ids, ticket ids, branch names, filesystem paths, model ids, raw trace detail.

One opt-in list in `app.css` owns the boundary, so the rule is enforceable instead of aspirational.

Also replaced **18 hand-written `ui-monospace, SFMono-Regular, …` stacks** with the token. Every one of them omitted JetBrains Mono, so "the mono face" was already three different faces depending on which surface you were looking at.

## Layout

- `.page-stack` is capped at **1120px and centred**; the settings detail pane at 720px. A repo row no longer puts its name and its buttons ~900px apart on a 1440px display.
- **Ticket rows** fold agent/model into the meta line instead of pinning them to the far right edge, hundreds of pixels from the title they describe.
- **Chat list rows** get real gutters, and a single nominated item absorbs overflow — so the meta line stops truncating all three values at once (`smoke-re… · TICKET-350-smoke-fixtu… · model unkno…` → `smoke-repo · TICKET-350-smoke-fixt… · model unknown`).

## Chat transcript

- **Bottom-anchored.** The virtualised list is `flex: 1 1 auto`, so the existing `margin-top: auto` never applied — a two-message chat sat ~800px above the composer.
- **The user's turn carries the accent tint**; the assistant's is bare prose with a quiet byline that drops on consecutive turns. Turns are a full line apart, at a 760px reading measure.
- **Live run state is a floating pill** instead of a grey bar ruled across the conversation.
- The **send button** is a real primary action, and reads as "nothing to send yet" rather than broken when disabled.

## Silent CSS failures fixed

`var(--x)` on an undeclared token **does not fall back** — the declaration is invalid at computed-value time and the property drops to inherited/initial. Nothing warns. Seven were live in `main`:

| Token | Symptom |
| --- | --- |
| `--color-surface-raised`, `--shadow-sm` | slash-command menu rendered **transparent** over the transcript |
| `--color-text`, `--radius-sm` | chat title lost its colour and radius |
| `--font-size--1` | agent status pill silently inherited body size |
| `--color-accent-strong`, `--ease-in-out` | hover colour and easing dropped |

`designTokens.test.ts` now fails the build on any such reference — and caught one of them itself while I was writing it.

## Also

- `.message span` and `.chat-row-meta` were **descendant**-scoped, so they restyled inline `<span>`s inside rendered markdown (uppercase 11px blocks mid-sentence). Both are now direct-child.
- List and summary previews strip markdown through a shared `markdownPreview()` — ticket rows no longer read `## Goal Fixture ticket used by \`scripts/x.py\`. ## Evidence — …`.
- Design-doc violations visible on repo/worktree detail: tinted zero KPIs (a green `0 active` reads as success, a red `0 needs fix` as alarm), a shouted `DANGER ZONE` eyebrow, heading-as-link with a trailing arrow, trailing-arrow ghost buttons.
- **`DESIGN.md`** documented an indigo/Inter system with a 4px spacing scale that had not matched the code for some time, and pointed at three components and helpers that no longer exist (`PmaMemoryView.svelte`, `.memory-tabs-v2`, `repoInitials()` in the wrong module). Rewritten against reality, plus the new typography, measure, and transcript rules.

## Unrelated pre-existing test failures, fixed to unblock the hook

All were already red on `main`; none involve code this branch changes.

- **`test_managed_thread_runtime.py` (×2)** asserted the runtime prompt ends with the literal user text. That stopped being true when the web surface began appending artifact-delivery instructions to the execution message. Repaired to assert what they were actually guarding: nothing is appended after the closing `</user_message>`, and the user's text opens the final block. Mutation-checked — injecting trailing content after the block still fails them.
- **`test_execution_history_maintenance.py` (×2)** and **`test_cli_hub_orchestration_history.py` (×1)** were a **time bomb**. The audit only flags a terminal execution as missing a manifest while it is inside the 90-day cold-trace retention window, and the fixture hardcoded `2026-04-12`. They passed until that date aged out of the window, then failed on a day nobody touched the code. Fixture timestamps are now anchored to "yesterday"; the deliberately-ancient `2000-01-01` cases stay literal. Mutation-checked — pushing the fixture back outside the window fails them again.

## Verification

`scripts/check.sh` passes end to end (this is what the pre-commit hook runs):

- 1008 vitest tests, `svelte-check` clean (665 files, 0 errors), eslint clean
- 9969 pytest tests, mypy strict repo-wide (1000 source files), black clean
- Playwright screenshot packs at 1440×1000 and 390×844, light and dark

Two known-unrelated items remain red and are **not** addressed here:

- `web_ui_smoke_journeys` ticket-detail — stuck on its loading skeleton; reproduces unchanged on the base commit `3999c723`.
- One process-supervisor test that flakes only under xdist contention — a different test each run, green every time in isolation.

Note for whoever picks this up: this worktree's venv was missing `pytest-xdist` and `hypothesis` and had `black` 25.11.0 against the pinned 26.5.1, so the Python lane had effectively not been running here. Installing the pinned dev deps is what surfaced the time-bomb tests above.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/codex_autorunner/adapters/discord/interaction_session.py
+++ b/src/codex_autorunner/adapters/discord/interaction_session.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import enum
 import logging
 from collections import OrderedDict
-from typing import Any, Optional
+from typing import Any, Optional, cast
 
 from .components import DISCORD_SELECT_OPTION_MAX_OPTIONS
 from .errors import DiscordAPIError, is_unknown_interaction_error
@@ -90,7 +90,7 @@ class DiscordInteractionSession:
             InteractionInitialResponse.DEFER_EPHEMERAL,
             InteractionInitialResponse.DEFER_COMPONENT_UPDATE,
         }:
-            return self.initial_response.value
+            return cast(str, self.initial_response.value)
         return None
 
     def _max_message_length(self) -> int:
--- a/src/codex_autorunner/web_frontend/DESIGN.md
+++ b/src/codex_autorunner/web_frontend/DESIGN.md
@@ -21,9 +21,17 @@ The hub is a focused, internal product for engineers. The visual taste is
 - **Cards do work.** Each row/card carries a clear identity (avatar, name),
   meta (branch, time, counts), and an action surface (hover chevron, click
   anywhere). Avoid double-nesting cards inside panels inside cards.
-- **Density without crowding.** Prefer `--space-2`/`--space-3` between
-  related elements, `--space-4`/`--space-5` between sections. Don't waste a
-  half-screen on a header that says "Repos".
+- **Density without crowding.** Prefer `--space-3`/`--space-4` between
+  related elements, `--space-5`/`--space-6` between sections. Don't waste a
+  half-screen on a header that says "Repos" — but don't crush a row into 4px
+  of padding either, so that its own text truncates against the column edge.
+  Density is about how much you show, not about how little air it gets.
+- **Prose is prose; machine text is machine text.** The proportional face
+  carries language, the mono face carries identifiers. Getting this backwards —
+  as this app did, by setting mono globally — is what makes an interface read
+  as generated rather than designed. See Typography.
+- **The reading column has a measure.** Content is centred and capped, not
+  stretched to whatever the display happens to be. See Measure.
 - **Hover earns affordance.** Borders strengthen, shadows lift gently,
   hidden chevrons slide in. Never animate layout-shifting properties.
 - **Each fact lives in exactly one place.** If the breadcrumb already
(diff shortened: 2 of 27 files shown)


Git-on-my-level:
@cursor address new review comment and fix failing CI

